Issue #124, opened 2021-04-12
`make distcheck` omits some files because their names are too long


smcv:
In a Steam Runtime 2 'soldier' SDK chroot (based on Debian 10):

```
...
(steamrt_soldier_amd64)user@host:~/SDL_ttf$ make distcheck
tardir=SDL2_ttf-2.0.15 && ${TAR-tar} chof - "$tardir" | eval GZIP= gzip --best -c >SDL2_ttf-2.0.15.tar.gz
tar: SDL2_ttf-2.0.15/Xcode/macOS/SDL2.framework/Versions/Current/Frameworks/hidapi.framework/Resources/LICENSE-bsd.txt: file name is too long (max 99); not dumped
tar: SDL2_ttf-2.0.15/Xcode/macOS/SDL2.framework/Versions/Current/Frameworks/hidapi.framework/Resources/AUTHORS.txt: file name is too long (max 99); not dumped
tar: SDL2_ttf-2.0.15/Xcode/macOS/SDL2.framework/Versions/Current/Frameworks/hidapi.framework/Resources/LICENSE-gpl3.txt: file name is too long (max 99); not dumped
tar: SDL2_ttf-2.0.15/Xcode/macOS/SDL2.framework/Versions/Current/Frameworks/hidapi.framework/Resources/LICENSE.txt: file name is too long (max 99); not dumped
tar: SDL2_ttf-2.0.15/Xcode/macOS/SDL2.framework/Versions/Current/Frameworks/hidapi.framework/Resources/LICENSE-orig.txt: file name is too long (max 99); not dumped
tar: SDL2_ttf-2.0.15/Xcode/macOS/SDL2.framework/Versions/Current/Frameworks/hidapi.framework/Resources/Info.plist: file name is too long (max 99); not dumped
tar: SDL2_ttf-2.0.15/Xcode/macOS/SDL2.framework/Versions/Current/Frameworks/hidapi.framework/Versions/Current/: file name is too long (max 99); not dumped
tar: SDL2_ttf-2.0.15/Xcode/macOS/SDL2.framework/Versions/Current/Frameworks/hidapi.framework/Versions/A/Resources/: file name is too long (max 99); not dumped
tar: SDL2_ttf-2.0.15/Xcode/macOS/SDL2.framework/Versions/Current/Frameworks/hidapi.framework/Versions/A/_CodeSignature/: file name is too long (max 99); not dumped
[...]

PR #125, opened 2021-04-12
build: Use POSIX 1003.1-2001 tar archives for `make dist`

smcv:
* build: Use the form of AC_INIT with arguments
    
    AC_INIT with no arguments is deprecated, and so is the 2-argument form
    of AM_INIT_AUTOMAKE (since 2002). The post-2002 equivalent is to
    pass the project name and the version number to AC_INIT, which defines
    VERSION from its argument.
    
    AC_INIT does not support the version number being expanded from shell
    variables, so convert it to be expanded from m4 macros.
    
    To preserve the tarball release names that SDL_ttf has traditionally
    used, we need to use the 4-argument form of AC_INIT. because by default
    Autoconf normalizes the name to lower-case. While we're doing that,
    we might as well use a non-empty third argument to specify the bug
    reporting address.

* build: Move Automake options into configure.ac
    
    Some Automake options, such as tar-ustar, can only be passed in this
    way, and cannot be in the $(AUTOMAKE_OPTIONS) make variable.

* build: Use POSIX 1003.1-2001 tar archives for `make dist`
    
    The Automake default is to use ancient UNIX v7 tar archives, which do
    not support long filenames, causing some of the macOS stuff to be
    excluded from `make dist`.
    
    Resolves: https://github.com/libsdl-org/SDL_ttf/issues/124

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,6 +1,5 @@
 # Makefile.am for the SDL truetype font loading library and viewer
 ACLOCAL_AMFLAGS = -I acinclude
-AUTOMAKE_OPTIONS= foreign subdir-objects
 
 lib_LTLIBRARIES = libSDL2_ttf.la
 
--- a/configure.ac
+++ b/configure.ac
@@ -45,7 +45,7 @@ dnl Detect the canonical build and host environments
 AC_CANONICAL_HOST
 
 dnl Setup for automake
-AM_INIT_AUTOMAKE([foreign subdir-objects])
+AM_INIT_AUTOMAKE([foreign subdir-objects tar-pax])
 
 dnl Check for tools
 AC_PROG_CC
--- a/configure.ac
+++ b/configure.ac
@@ -45,7 +45,7 @@ dnl Detect the canonical build and host environments
 AC_CANONICAL_HOST
 
 dnl Setup for automake
-AM_INIT_AUTOMAKE
+AM_INIT_AUTOMAKE([foreign subdir-objects])
 
 dnl Check for tools
 AC_PROG_CC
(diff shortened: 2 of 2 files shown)


sezero:
Thanks. Will have to port this to other projects like SDL_image, soon.

smcv:
As noted on #124, if `make distcheck` succeeds with `tar-ustar` in other projects, that might be *slightly* more portable - I only used `tar-pax` here because when I tried `tar-ustar`, it turned out not to work.

sezero:
One issue is that `make dist` doesn't store symlinks as symlinks but follows them.
That actually seems to be the case without this patch too, so it should be something
with automake.

sezero:
Made the same changes in SDL_image as https://github.com/libsdl-org/SDL_image/commit/a291357d0b88884c5610e848d02a21ee26f405c7

@smcv: Do you know a solution so that `make dist` stores symlinks as they are without following them?

smcv:
> Do you know a solution so that make dist stores symlinks as they are without following them?

I think it might be deliberate that `make dist` converts symlinks into copies, so that the tar archive can be unpacked on OSs with a more limited filesystem interface, like Windows.

If it's important that symlinks stay as symlinks, one option is for official releases to be a `git archive` instead of a `make dist`, and another option would be to delete the copy and replace it with a symlink in a `dist-hook`?

The context for this MR is that I was asked to package a snapshot of `SDL_ttf`, so I wanted to start from a source code archive that was mostly equivalent (except for version number, etc.) to what would be in the official release if `SDL_ttf` was released today. Trying to use a `make dist` was just a guess - it isn't obvious to me at the moment how official releases are actually made.

PR #126, opened 2021-05-02
CMake: Support static-only SDL2

glebm:
When SDL2 is built with `-DBUILD_SHARED_LIBS=OFF`, it does not define an `SDL2::SDL2` target.
    
Use `SDL2::SDL2-static` and set `POSITION_INDEPENDENT_CODE` if `BUILD_SHARED_LIBS` is `OFF`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -57,7 +57,13 @@ set(SDL_TTF_VERSION "${SDL_TTF_MAJOR_VERSION}.${SDL_TTF_MINOR_VERSION}.${SDL_TTF
 
 ##### library generation #####
 add_library(SDL2_ttf SDL_ttf.c SDL_ttf.h)
-target_link_libraries(SDL2_ttf SDL2::SDL2 Freetype::Freetype)
+if (BUILD_SHARED_LIBS)
+  target_link_libraries(SDL2_ttf SDL2::SDL2)
+else ()
+  target_link_libraries(SDL2_ttf SDL2::SDL2-static)
+  set_target_properties(SDL2_ttf PROPERTIES POSITION_INDEPENDENT_CODE ON)
+endif ()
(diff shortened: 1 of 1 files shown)


1bsyl:
Hi @sezero, should we need this ?
why isn't this needed in other libs: SDL_image and SDL_mixer ?

sezero:
First off: I'm not good with cmake (and I don't like cmake, I detest cmake),
so, as far as cmake features are concerned I may be wrong.

That aside, why do we need PIC with static builds?

As for cmake in SDL_mixer: it is very much incomplete (IIRC targeting only Android),
so having this in SDL_mixer should be the least of concerns.

sezero:
Also, why can we not support both static and dynamic builds in one shot with cmake, like autotools?

1bsyl:
Ok, I was just wondering, I am not a cmake expert neither, and I don't use it for SDL. It seems that all the sdl cmake files need some re-engineering.

glebm:
@1bsyl

> why isn't this needed in other libs: SDL_image and SDL_mixer ?

SDL_image does this already:

https://github.com/libsdl-org/SDL_image/blob/d026716b63e761f3fc9f64725480cf37a9dac81f/CMakeLists.txt#L74-L78

https://github.com/libsdl-org/SDL_image/blob/d026716b63e761f3fc9f64725480cf37a9dac81f/CMakeLists.txt#L14-L16

glebm:
> That aside, why do we need PIC with static builds?

If you want to link shared libraries against static SDL_ttf, it needs to be PIC

1bsyl:
@glebm, this is true, that was added recently here https://github.com/libsdl-org/SDL_image/commit/f51f5f696abd75c82e12a527a4d8deb9c2ae9cc5
( I grepped on an older version )

glebm:
@1bsyl Thanks Sylvain!

PR #127, opened 2021-05-02
CMake: Set POSITION_INDEPENDENT_CODE on static lib

glebm:
If building statically, set `POSITION_INDEPENDENT_CODE ON` to pass `-fPIC`.


glebm:
Added this part of #126 instead

PR #129, opened 2021-05-22
add relative include path for SDL in Xcode project

lufinkey:
I previously had my project set up with an SDL_ttf submodule pointing to the Xcode-iOS project. The new Xcode project doesn't automatically include SDL when the submodules are placed in the same folder. This change corrects that.

Note: I tried including it in the normal "Header Search Paths", but it refused to work. It only worked as a user header search path

### DIFF
--- a/Xcode/SDL_ttf.xcodeproj/project.pbxproj
+++ b/Xcode/SDL_ttf.xcodeproj/project.pbxproj
@@ -2000,7 +2000,7 @@
 				PRODUCT_BUNDLE_IDENTIFIER = "org.libsdl.SDL2-ttf";
 				PRODUCT_NAME = SDL2_ttf;
 				TVOS_DEPLOYMENT_TARGET = 9.0;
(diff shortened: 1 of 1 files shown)


1bsyl:
@lufinkey 
not sure, but maybe this should be set to:

$(HOME)/Library/Frameworks/SDL2.framework/Headers /Library/Frameworks/SDL2.framework/Headers

like in https://github.com/libsdl-org/SDL_image/blob/main/Xcode/SDL_image.xcodeproj/project.pbxproj#L504

lufinkey:
The problem is that this doesn't work when you're building from the git
repos. I have all the SDL libraries as submodules, and this fails to
recognize the main SDL submodule, even though it was able to do it fine in
the other Xcode project.

lufinkey:
You need the relative include path

1bsyl:
Ok, it also seem not building currently for me but only for the iphonesimulator version of libs.
and this was present in older version of xcode project

lufinkey:
That's odd, IPhone simulator was what I was targeting and this fixed it for
me.

1bsyl:
now, all builds correctly

1bsyl:
for the record, this was present here:
https://github.com/libsdl-org/SDL_ttf/commit/9000cd7a2c91a087a4265957874490025e32c958#diff-9756f3169d91c3a7135817bcd67535ae716f9616aa232ac03f2b142cd6a6822bL583

Issue #131, opened 2021-05-29
ClearType-style LCD rendering


andre-barata:
This is a request, if possible, to add support for ClearType-style LCD rendering, which FreeType already seems to [support](https://www.freetype.org/freetype2/docs/reference/ft2-lcd_rendering.html).

This makes smaller font sizes more readable, by basically using the R G and B portions of every pixel as additional subpixels with [different luminance levels](https://en.wikipedia.org/wiki/Subpixel_rendering). 
![image](https://user-images.githubusercontent.com/6938680/120062621-0dbd5180-c05b-11eb-817e-40f4c9d70051.png)
 
maybe a new hinting type would be a good place to set the feature on. like :
`TTF_SetFontHinting(mainFont, TTF_HINTING_LCD_SUBPIXEL);`

1bsyl:
Did you try the subpixel rendering mode which is present in head source ?
https://github.com/libsdl-org/SDL_ttf/blob/main/SDL_ttf.h#L141

which lcd hardware do you use ?

andre-barata:
Yes, I'm currently using that mode on a [project](https://github.com/andre-barata/GridMerge/tree/3f91f46d9f9b38ea6ba9192ba9436428c9340189) with the latest version of SDL_ttf.
I'm using the new hint type `TTF_SetFontHinting(mainFont, TTF_HINTING_LIGHT_SUBPIXEL);`. 
The readability is better, but still not as good as with the colored edges on the fonts. here's an example:

![image](https://user-images.githubusercontent.com/6938680/120085395-b0171c80-c0cf-11eb-81b3-d362b8a41f21.png)

1bsyl:
I give a try and this seems almost feasible to add this rendering.
but that would only make sense with SDL *blended* function which returns an ARGB surface.

the thing is that blended, also require an argb color parameter 
https://github.com/libsdl-org/SDL_ttf/blob/main/SDL_ttf.h#L284
and I am not sure how to blend the RGB glyph with this argb color parameter, any idea?

1bsyl:
Hi @andre-barata 
Here's an archive, can you give a try. [LCD.zip](https://github.com/libsdl-org/SDL_ttf/files/6597997/LCD.zip)
Either patch your SDL_ttf with the .diff file, or re-use the SDL_ttf.c/h files.

You need to set the hinting to LCD, use the TTF_*Blended functions to render text.

Still not sure how the blended color should be used. 
[...]

PR #138, opened 2021-07-05
Add functions to use FreeType ClearType-style LCD rendering

1bsyl:
As proposed in issue #131, add functions to use FreeType ClearType-style LCD rendering

### DIFF
--- a/SDL_ttf.c
+++ b/SDL_ttf.c
@@ -168,11 +168,13 @@ static SDL_INLINE int hasNEON()
 #define DIVIDE_BY_255(x)    DIVIDE_BY_255_SIGNED(x, 1)
 
 
-#define CACHED_METRICS  0x10
+#define CACHED_METRICS  0x20
+
 #define CACHED_BITMAP   0x01
 #define CACHED_PIXMAP   0x02
 #define CACHED_COLOR    0x04
-#define CACHED_SUBPIX   0x08
+#define CACHED_LCD      0x08
+#define CACHED_SUBPIX   0x10
 
 
 typedef struct {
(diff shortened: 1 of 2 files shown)


1bsyl:
We may want to define  "FT_CONFIG_OPTION_SUBPIXEL_RENDERING"
external/freetype-2.10.4/include/freetype/config/ftoption.h

```
    * Uncomment the line below if you want to activate LCD rendering
    * technology similar to ClearType in this build of the library.  This
    * technology triples the resolution in the direction color subpixels.  To
    * mitigate color fringes inherent to this technology, you also need to
    * explicitly set up LCD filtering.
    *
    * When this macro is not defined, FreeType offers alternative LCD
    * rendering technology that produces excellent output.
```

adamharrison:
Just wanted to say, this is great. I'd love to use this when it gets merged!

manuel-fischer:
It might be useful if there would be transparency support.
But for correct LCD rendering with transparency support, SDL itself should add support for separate alpha channels for red, green and blue, which might not be supported with hardware acceleration. 

It might also be useful, if the direction of the subpixels can be configured.

SDL_ttf.c
```diff
@@ -1449,6 +1548,68 @@ static SDL_Surface *Create_Surface_Blended(int width, int height, SDL_Color fg,
...
+
+        textbuf = SDL_CreateRGBSurfaceWithFormatFrom(pixels, width, height, 0, pitch, SDL_PIXELFORMAT_ARGB8888);
+        if (textbuf == NULL) {
+            SDL_free(ptr);
+            return NULL;
+        }
+
+        /* Let SDL handle the memory allocation */
+        textbuf->flags &= ~SDL_PREALLOC;
+        textbuf->flags |= SDL_SIMD_ALIGNED;
```
slouken, line 1598: This should only be set if the surface is allocated with SDL_SIMDAlloc(), otherwise it can cause a crash.

slouken: Ah, but I see you abuse the way the SIMD memory allocation works so that this is done correctly. Maybe we should add a function `SDL_SIMDAllocAligned(len, align)`?

Issue #157, opened 2022-01-03
Behavior of "Wrapped" with small lengths


Starbuck5:
I posted this under the SDL_ttf pre release announcement on [Discourse](https://discourse.libsdl.org/t/sdl2-ttf-2-0-18-prerelease/34240), but I also wanted to raise it here. Especially since this behavior will need to be supported as is after a release, for backwards compatibility.

I’ve been playing around with the wrapped text functions, and I’m concerned about how they handle small wrap values.

At wrap length 0, the text renders as a single line, and newline characters are still rendered as unknown characters. I expected this to behave like the normal text rendering functions but newline aware.

With a positive but very small wrap length, the text doesn’t fully show up. Or parts of letters might show up. I expected this to allocate at least enough space to put at least one character of the input string on each line.

Pinging the author of the functions @1bsyl. Thank you for all the work keeping SDL_ttf golden!

1bsyl:
@Starbuck5 
wrapLength 0 seems a good idea. 
Please give a try to this PR : https://github.com/libsdl-org/SDL_ttf/pull/158

> With a positive but very small wrap length, the text doesn’t fully show up. Or parts of letters might show up. I expected this to allocate at least enough space to put at least one character of the input string on each line.

We don't create a surface with more than wrapLength, so I am not sure

Starbuck5:
Thanks for such a quick reply and PR!

I've gotten some infrastructure set up to actually build SDL_ttf locally, so I'll be able to test that tomorrow.

I figure I should provide some examples of what I'm talking about with the small wrap length weirdness.

This is using the string "hello world, how bout that?", and the font freesansbold.ttf:
At wrap length 11 (just the first char):
![11](https://user-images.githubusercontent.com/46412508/148035325-e0eba926-7412-4de1-8144-ddf7d8884b00.png)
At wrap length 12 (stuff is missing off the end):
![12](https://user-images.githubusercontent.com/46412508/148035366-4ac1ae95-7b66-4c72-82a2-56c02aad638f.png)
At wrap length 15 (a good result):
![15](https://user-images.githubusercontent.com/46412508/148035430-4f99fe0d-3cbc-4120-974e-8760830c4ca7.png)

> We don't create a surface with more than wrapLength, so I am not sure

I think it would be okay to, if needed.

slouken:
No, the wrap length is a hard limit. You should ensure that your UI areas are large enough for at least one character if you want one character to be visible.

[...]

PR #158, opened 2022-01-03
If wrapLength is 0, only wrap on new lines.

1bsyl:
At wrap length 0, the text is cut at each new lines ('\n')

See https://github.com/libsdl-org/SDL_ttf/issues/157

### DIFF
--- a/SDL_ttf.c
+++ b/SDL_ttf.c
@@ -3364,7 +3364,11 @@ static SDL_Surface* TTF_Render_Wrapped_Internal(TTF_Font *font, const char *text
 
             if (numLines >= maxNumLines) {
                 char **saved = strLines;
-                maxNumLines += (width / (wrapLength + 1)) + 1;
+                if (wrapLength == 0) {
+                    maxNumLines += 32;
+                } else {
+                    maxNumLines += (width / wrapLength) + 1;
+                }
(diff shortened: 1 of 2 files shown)


SDL_ttf.c
```diff
@@ -3350,7 +3364,7 @@ static SDL_Surface* TTF_Render_Wrapped_Internal(TTF_Font *font, const char *text
 
             if (numLines >= maxNumLines) {
                 char **saved = strLines;
-                maxNumLines += (width / wrapLength) + 1;
+                maxNumLines += (width / (wrapLength + 1)) + 1;
```
slouken, line 3367: Why are you making this change?

1bsyl: if wrapLength is zero, then it would divide by 0 and crash, isn'it ?

slouken: Yes, but it actually changes the maxNumLines calculation for certain values of width and wrapLength. Maybe instead you can explicitly check for wrapLength == 0?

1bsyl: I don't think it matters, it is simply the increment for the memory in to realloc, so that we don't realloc for each new line pointers.

but, since we allow wrapLength == 0; maybe a we could just use a fixed increment. so that on don't increment of "width" in one shot.

Issue #160, opened 2022-01-07
Add text align and line spacing to SDL_ttf?


Starbuck5:
With the new Wrapped functions adding automatic line wrapping, it would be cool to get support for line spacing and text alignment as well.

I've got it working locally (hardcoded), and the code in the `wrapped` functions makes it really easy to integrate these features.

This could handle aligns if wired together properly (comments are what I added) (around line 3515 in the current source)
```c
        /* right align */
        //xstart += wrapLength - line_width;

        /* center align */
        //xstart += wrapLength / 2 - line_width / 2;

        /* Render one text line to textbuf at (xstart, ystart) */
        if (Render_Line(render_mode, font->render_subpixel, font, textbuf, xstart, ystart, fg.a) < 0) {
            goto failure;
        }
```

For text spacing, it seems to work fine just multiplying lineskip (on ~line=3432) by some scalar. For line spacing, I've been imagining it as a scale, not as an absolute number of pixels. So lineskip=1 is default, lineskip=1.5 puts more space in between each line, etc.

This might be burdensome to add to the `Wrapped` functions as they are now because users would have to specify these arguments, which would be inconvenient to those who just want to throw in some simple multi line text. So instead, perhaps it could be added as a series of `WrappedEx` functions.

I'd be happy to submit a PR for this.

@1bsyl what do you think? (sorry to keep pestering you about line wrapping features 😄)

1bsyl:
We could add new function like TTF_SetLineSkip()  and TTF_SetLineAlignement() for instance to set some style configuration.
I think you can submit a PR for that.

It's not possible to change the wrapped prototype !
And it doesn't seem reasonable to add a new wrapped function for this.

If we want to add a more complex wrapped function, we may want more things: like bidi, multiples ttf arguments, more styles maybe ...

slouken:
I'm wondering if maybe we want to create a "meta" font rendering function that takes text and a parameter struct that has extended options. What do you think, @1bsyl?

1bsyl:
If it's only alignment and lineSkip, the usual 1 api seems fine. And it's already what we have for other settings.
I don't think we will add much more.
But we could even build meta struct interface on top of those function, so can get/set the current style

The Wrapped API are anyway limited if people want to mix various things inside color/style/font etc.
If one want something more complex or specific, it is built top of the Render functions.

slouken:
https://github.com/libsdl-org/SDL_ttf/pull/161 is merged!

PR #161, opened 2022-01-10
Add left/right/center align support to wrapped text

Starbuck5:
Hello. This covers part of #160.

I probably did plenty of stuff wrong.

The challenging bit was getting the underlines and strikethroughs to move with the text, so I added another parameter to drawLine that moves everything over on the x.

### DIFF
--- a/SDL_ttf.c
+++ b/SDL_ttf.c
@@ -294,6 +294,9 @@ struct _TTF_Font {
     hb_direction_t hb_direction;
 #endif
     int render_sdf;
+
+    /* Extra layout setting for wrapped text */
+    int horizontal_align;
 };
 
 /* Tell if SDL_ttf has to handle the style */
@@ -994,10 +997,11 @@ static SDL_INLINE void BG_NEON(const TTF_Image *image, Uint8 *destination, Sint3
 #endif
 
 /* Underline and Strikethrough style. Draw a line at the given row. */
-static void Draw_Line(TTF_Font *font, const SDL_Surface *textbuf, int row, int line_width, int line_thickness, Uint32 color, const render_mode_t render_mode)
+static void Draw_Line(TTF_Font *font, const SDL_Surface *textbuf, int column, int row, int line_width, int line_thickness, Uint32 color, const render_mode_t render_mode)
 {
     int tmp    = row + line_thickness - textbuf->h;
-    Uint8 *dst = (Uint8 *)textbuf->pixels + row * textbuf->pitch;
+    int x_offset = column * textbuf->format->BytesPerPixel;
+    Uint8 *dst = (Uint8 *)textbuf->pixels + row * textbuf->pitch + x_offset;
 
 #if TTF_USE_HARFBUZZ
     /* No Underline/Strikethrough style if direction is vertical */
@@ -3451,11 +3455,11 @@ static SDL_Surface* TTF_Render_Internal(TTF_Font *font, const char *text, const 
 
     /* Apply underline or strikethrough style, if needed */
     if (TTF_HANDLE_STYLE_UNDERLINE(font)) {
-        Draw_Line(font, textbuf, ystart + font->underline_top_row, width, font->line_thickness, color, render_mode);
+        Draw_Line(font, textbuf, 0, ystart + font->underline_top_row, width, font->line_thickness, color, render_mode);
     }
 
     if (TTF_HANDLE_STYLE_STRIKETHROUGH(font)) {
-        Draw_Line(font, textbuf, ystart + font->strikethrough_top_row, width, font->line_thickness, color, render_mode);
+        Draw_Line(font, textbuf, 0, ystart + font->strikethrough_top_row, width, font->line_thickness, color, render_mode);
     }
 
     if (utf8_alloc) {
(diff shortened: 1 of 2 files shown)


1bsyl:
I haven't tried it, but at first, it looks neat. 
just: 
- remove the "printf("setting font align=%i\n", align);"
- align the "if else" condition

please test with valgrind, strike / underline and Blended / Shaded / Mono

Starbuck5:
> remove the "printf("setting font align=%i\n", align);"

Ah, I hoped I could force push fast enough that no one would notice. 😅

I get what you mean with the `if/else` align.

I might have trouble with valgrind on this system, but this should be good with memory because I didn't do any mallocs.

1bsyl:
@Starbuck5 just re-align the "if else if" but that's not a big deal
I am going to try the patch

sezero:
Removing the extra comment from the commit message would be good.

Starbuck5:
> Removing the extra comment from the commit message would be good.

Oh sorry, that was an artifact of my squash. Not intentional.

1bsyl:
It seems to me it crashes when rendering a corner case.

Actually this is an "Invalid write of size 4"

This happen in Draw_Line (
SDL_memset4 (SDL_stdinc.h:453)
==24065==    by 0x4A96084: Draw_Line (SDL_ttf.c:927) )

align: right
wrappedLenght: 661 (eg big)
string:   "="
Blended
font: /system/fonts_AndroidN_Preview/DancingScript-Regular.ttf

I cannot provide a real testcase, I use a test app. with fuzzing. [SDL_ttf_testapp.c.txt](https://github.com/libsdl-org/SDL_ttf/files/7838253/SDL_ttf_testapp.c.txt)

(you need to fill with font path insides), 
F1= help
press 'r' to do random rendering
( comment out the line 615, this is to 'replay' a crash configuration)

I may have a closer look, but not right now!

1bsyl:
@Starbuck5 
I looked again, and we need to change 'wrapLength' to 'width'
because if the surface won't be wider that wrapLength for sure, it may be smaller!
```
        if (font->horizontal_align == TTF_ALIGN_RIGHT) {
            xoffset = width - line_width;
        } else if (font->horizontal_align == TTF_ALIGN_CENTER) {
            xoffset = width / 2 - line_width / 2;
        } else {
            xoffset = 0;
        }
```
I'll run the fuzzing in background to detect any other issue ... 

Otherwise, maybe we should rename the function so that it's clear the alignment is for the Wrapped methods ? 
Any suggestion @slouken ?

Starbuck5:
> surface won't be wider that wrapLength for sure, it may be smaller!

Ah, I was assuming it was guaranteed at the wrapLength width.

For API, maybe it should be `TTF_SetFontWrappedAlign`.

Starbuck5:
> because if the surface won't be wider that wrapLength for sure, it may be smaller!

Oh, this might be a problem for right and center aligned text. For example, if it's right aligned at wrapLength=500, it should end at the 500th pixel, not the last pixel of whatever the longest width was.

1bsyl:
> > because if the surface won't be wider that wrapLength for sure, it may be smaller!
> 
> Oh, this might be a problem for right and center aligned text. For example, if it's right aligned at wrapLength=500, it should end at the 500th pixel, not the last pixel of whatever the longest width was.

yes, then add: 

line ~3510

```patch
             /* Don't go above wrapLength if you have only 1 line which hasn't been cut */
             if (font->horizontal_align == TTF_ALIGN_RIGHT) {
                 width = SDL_min((int)wrapLength, width);
             }
+            if (font->horizontal_align != TTF_ALIGN_RIGHT) {
+                width = wrapLength;
+            }
```

so that it creates a wrapLength surface

1bsyl:
for me it passes the valgrind/fuzzing testing with 100 000 wrapped string with no issue detected.

Starbuck5:
> for me it passes the valgrind/fuzzing testing with 100 000 wrapped strings with no issue detected.

Nice!

I pushed some more changes. One of which was directly suggested by you. The other one I changed up a bit.

Since the surface can only be less than wrapLength if it's a single line, and only if it's ALIGN_LEFT, it seemed simpler to have it always be wrapLength wide. Maybe that would also be convenient to people calling the function.

New problem I found is that it no longer handles wrapLength = 0 as it used to. I don't think it's from this latest commit.

1bsyl:
Indeed I fixed the issue with wraplength = 0;
https://github.com/libsdl-org/SDL_ttf/commit/3e702ed9bf400b0a72534f144b8bec46ee0416cb

Please keep the legacy behavior so that the surface may be < wrapLength if align is left

Starbuck5:
> Please keep the legacy behavior so that the surface may be < wrapLength if align is left

Alrighty.

@1bsyl I believe this is now logically equivalent to how it was before. It can only be less than wrapLength if it's 1 line and if it's left aligned.

1bsyl:
Yes, this seems ok to me !
just the function name, and if we want to add a function to store a meta style struct as @slouken suggested

Starbuck5:
Alright, I'll change the name to `TTF_SetFontWrappedAlign`, if that's alright. It's more descriptive than what I've got in now.

The funny thing about the struct idea is that I'm implementing this so I can use it downstream in Pygame, and I was planning to expose that as an extra "layout" dictionary argument.

But on this level, with just a couple of them, I think it would be simpler (for us and for the users) to do it with functions as I've done here.

1bsyl:
@Starbuck5 let's wait for @slouken last word so you don't change back and forth multiple times :)

1bsyl:
Don't worry, we can "squash and merge" in the end

Starbuck5:
Definitely!

If we're going to go down the `TTF_SetFontWrappedAlign` path, maybe the constants should be named like `TTF_WRAPPED_ALIGN_LEFT`.

Starbuck5:
@slouken Waiting on you for an executive decision on the API 🙂

slouken:
> If we're going to go down the `TTF_SetFontWrappedAlign` path, maybe the constants should be named like `TTF_WRAPPED_ALIGN_LEFT`.

Yep, this renaming sounds good.

PR #166, opened 2022-01-11
Document which functions clear the glyph cache

albertvaka:
I think there's some value in mentioning this, since it has an impact on performance if functions like `TTF_SetFontSize` are used frequently. 

The wording might not be the best, feel free to reword the comment.


1bsyl:
This seems ok to me !

1bsyl:
Thanks !

PR #170, opened 2022-01-14
Make cmake build work on PSP

sharkwouter:
The PSP does not support using the ``POSITION_INDEPENDENT_CODE`` property.


sezero:
That `POSITION_INDEPENDENT_CODE` crapola needs optionalizing, at the least,
if not removed...  That said, if someone wants to merge this, do so - the cmake support
in here is immature yet.

sharkwouter:
What are the problems with the CMake support? For the PSP it seems to result in a working binary and ``make install`` seems to work like expected. I've not really seen a difference.

sezero:
> What are the problems with the CMake support

See the FIXME notes at top of CMakeLists.txt.

Issue #175, opened 2022-01-27
TTF_SetDirection and TTF_SetScript both require HarfBuzz constants/headers


a-hurst:
Hi there,

With the new 2.0.18 release, TTF added functions that allow you to set the text direction and script type for the underlying HarfBuzz library. This is great, except that SetDirection requires HarfBuzz constants as an input and SetScript requires a value from the HarfBuzz `HB_TAG` macro as an input, making it difficult to use these functions without also importing the HarfBuzz headers.

Given that the SetDirection constants and `HB_TAG` macro are pretty simple, would you consider adding TTF-specific alternatives to them for these functions? e.g `TTF_DIRECTION_LTR` instead of `HB_DIRECTION_LTR`, `TTF_SCRIPT` instead of `HB_TAG`, etc. That would help make TTF fully self-contained, and also make it easier to document down the road.

Thanks in advance!

slouken:
Yes, that seems like a good idea, thanks!

1bsyl:
@a-hurst @slouken 
Here's some type that could be used in place of hb types to simply compilation
@a-hurst can you try ?

a-hurst:
@1bsyl Great! A couple questions/comments:

1. Why are the top-to-bottom and bottom-to-top direction constants commented out? Just not something TTF officially supports?
2. How would one go about testing the TTF\_SetScript function? I'm guessing you'd need an Arabic font or something and try setting the script to see if it renders any different? I don't actually have an immediate use case for either function, I just noticed the HarfBuzz constant/macro dependency while updating the PySDL2 bindings and thought it was worth mentioning.
3. To avoid the maintenance burden of all the `TTF_SCRIPT_` constants, I think it's fine as an end-user  to just have the `TTF_TAG` macro that converts the standardized ISO 15924 4-character codes to SetScript's required input format. All the possible 4-letter codes for different script styles are [already documented on unicode.org(https://unicode.org/iso15924/iso15924-codes.html), the only issue I can think of is TTF's HarfBuzz not supporting a newly-added ISO script tag and having no way to check for an error if it isn't available (though the lack of error checking is a problem with the current release as well).

1bsyl:
So this was added in: https://github.com/libsdl-org/SDL_ttf/commit/6dcf9e6f2341a482cc6461b885c5063b49096787

` int TTF_SetFontScriptName(TTF_Font *font, const char *script) `

PR #176, opened 2022-01-28
Add type exactly mapped to hb types (see #175)

1bsyl:
Add type exactly mapped to hb types (see #175)


sezero:
The idea is OK, but this needs continuous updates from harfbuzz mainstream.
And, which enum values are we restricting ourselves to? E.g.: ttf_script_t
added here is missing values present in hb >= 3.0.0.

Also why are `_TTB` and `_BTT` directions are commented out?

1bsyl:
- we don't do support _TTB _BTT as there are vertical direction.

- this is the value from our external/harfbuzz-2.8.0/ 
-  indeed, there not the one from newer lib

we could add a real ttf_script_t + ttf_direct_t but that's somehow also an api breaking. or is this ok ?

sezero:
>   this is the value from our external/harfbuzz-2.8.0/

The thing is, distros can and do build with `--disable-harfbuzz-builtin`, so they
are not restricted to our embedded version of harfbuzz.

1bsyl:
I've updated with real own TTF type

just need to add more hb_script and defines with HB version so that it compiles everywhere

sezero:
It now changes api, do we want that?

1bsyl:
It changes the API but it looks cleaner this way. ttf 2.0.18 is still new so it doesn't seem so an issue if 2.0.20 comes soon.
We can also revert to the previous version .. 
@slouken ?

sezero:
If we really do change api, it would be better to enumerate `TTF_SCRIPT_INVALID` as `-1`, so that we can insert more new values before it later.

slouken:
I'm thinking we should not change the API. If a new version of HarfBuzz comes out the application can pass through new values that SDL_ttf doesn't know about (and shouldn't care about.)

Don't change this yet, let's think this through a little more.

1bsyl:
Indeed TTF_SCRIPT_INVALID would be -1.

I think there are already lot of languages added: ~150 scripts. 1 script cover more than 1 language.
They won't add a huge amount of it.

Eventually, we could change the 'default' of the switch, so that if we don't recognize it as an TTF_Script, we just forward it to HB as  a raw value. Seem possible since TTF_Script doesn't seem to overwrite HB_TAG (4 chars)
 
so I wait, just ping me when we choose one solution.

a-hurst:
Reposting my relevant thoughts from #175 here since this seems like where the discussion is happening:

>To avoid the maintenance burden of all the `TTF_SCRIPT_` constants, I think it's fine as an end-user  to just have the `TTF_TAG` macro that converts the standardized ISO 15924 4-character codes to SetScript's required input format. All the possible 4-letter codes for different script styles are [already documented on unicode.org](https://unicode.org/iso15924/iso15924-codes.html), the only issue I can think of is TTF's HarfBuzz not supporting a newly-added ISO script tag and having no way to check for an error if it isn't available (though the lack of error checking is a problem with the current release as well).

@sezero @slouken does that solution sound more workable?

Also @1bsyl I haven't actually done a visual test yet so it could be totally broken, but my unit test for `TTF_SetDirection` in PySDL2 seems to show top-to-bottom rendering actually working:

```python
def test_TTF_SetDirection(font):
    # Define HarfBuzz direction constants
    HB_DIRECTION_LTR = 4
    HB_DIRECTION_TTB = 6

    # Try setting the script directiony 
    ret = sdlttf.TTF_SetDirection(HB_DIRECTION_LTR)
    assert ret == 0

    # Try changing the script direction to see if it has any effect
    w1, h1, w2, h2 = c_int(0), c_int(0), c_int(0), c_int(0)
    sdlttf.TTF_SizeText(font, b"Hi there!", byref(w1), byref(h1))
    ret = sdlttf.TTF_SetDirection(HB_DIRECTION_TTB)
    assert ret == 0
    sdlttf.TTF_SizeText(font, b"Hi there!", byref(w2), byref(h2))
    sdlttf.TTF_SetDirection(HB_DIRECTION_LTR) # Reset direction
    assert w1.value > w2.value
    assert h1.value < h2.value
```
Assuming it actually does what it's supposed to, is there any reason not to allow people to pass through those HarfBuzz values even if they're not officially supported?

1bsyl:
Vertical TTB and BTT aren't working currently.
but maybe it can be done ... needs to change horiAdvance, vertAdvance  and update the y position with it ...

1bsyl:
@a-hurst now this should be working to draw in vertical direction: https://github.com/libsdl-org/SDL_ttf/commit/119abf2e8958f7e73e67c3d0534115ad3eb6e167

1bsyl:
@slouken @sezero 
maybe that the function
```
extern DECLSPEC int SDLCALL TTF_SetDirection(int direction); /* hb_direction_t */
extern DECLSPEC int SDLCALL TTF_SetScript(int script); /* hb_script_t */
```
Should take the `TTF_Font *font` parameter in ? but that would change the API ...

sezero:
> @slouken @sezero maybe that the function
> 
> ```
> extern DECLSPEC int SDLCALL TTF_SetDirection(int direction); /* hb_direction_t */
> extern DECLSPEC int SDLCALL TTF_SetScript(int script); /* hb_script_t */
> ```
> 
> Should take the `TTF_Font *font` parameter in ? but that would change the API ...

Personally, I don't like changing the api, therefore I'd suggest adding a comment to the existing
functions like `These functions set/change the direction and script globally within SDL_ttf`
and adding new functions like `TTF_FontSetDirection` and `TTF_FontSetScript`, instead.

As for script and direction enums, I'd suggest an easy and ugly way like adding a note to
the top of our header -- something like:
```diff
diff --git a/SDL_ttf.h b/SDL_ttf.h
index 468359f..1122334 100644
--- a/SDL_ttf.h
+++ b/SDL_ttf.h
@@ -29,8 +29,14 @@
    confusion. */
 
+/* Note: In order to use direction and script for text shaping, define
+   SDL_TTF_HAVE_HARFBUZZ in your program before including SDL_ttf.h  */
+
 #ifndef SDL_TTF_H_
 #define SDL_TTF_H_
 
 #include "SDL.h"
+#ifdef SDL_TTF_HAVE_HARFBUZZ
+#include <hb.h>
+#endif
 #include "begin_code.h"
 
```

1bsyl:
@sezero I've added TTF_FontSetDirection() and TTF_FontSetScript()   (FontSet* vs SetFont to match current naming).  https://github.com/libsdl-org/SDL_ttf/commit/1b2f2d7b142ac090519b03a8213977e091f724dc

I agree to generally not to change the API. Especially if the api is older, with lot of dependencies. But for a very recent API, I think it should not be so strict. Even biggest project have the notion of 'deprecated' api.

About the script and direction enums. It's not about simply writing "#include <hb.h>" to have the hb type.
but configuring a project (android, cmake, linux, visual, xcode, etc..) and get harfbuzz header included for it. 
Not big deal, but more convenient if one has not to do so.

sezero:
> @sezero I've added TTF_FontSetDirection() and TTF_FontSetScript() (FontSet* vs SetFont to match current naming). [1b2f2d7](https://github.com/libsdl-org/SDL_ttf/commit/1b2f2d7b142ac090519b03a8213977e091f724dc)

`Draw_Line` might need adjusting? It currently uses the global state instead of the per-font direction.

khaledhosny:
Instead of using int, you can use strings and use the relevant _from_string() API in HarfBuzz.

1bsyl:
@sezero thanks fixed https://github.com/libsdl-org/SDL_ttf/commit/9a2cb0e452a52045419c3554e4c6696a3cd0a714, your reviewing is better than my test-cases

@khaledhosny thanks ! indeed, that may the best solution

slouken:
After thinking about it some more, we shouldn't add a script enumeration to our header, instead we should add a function that sets the script by name. People who want to use the enumeration can include the harfbuzz header themselves, and people who want to look it up by name can do so:
```c
extern DECLSPEC int SDLCALL TTF_SetFontScriptName(TTF_Font *font, const char *script);
```

1bsyl:
@slouken 
I have reverted not to change the old deprecated API (TTF_SetFont, TTF_SetScript).
Exposed also vertical direction TopToBottom and BottomToTop.
Change to TTF_SetFontScriptName
Doing some test seems that's ok.

if you merge, please 'squash' so we have only 1 commit

slouken:
It looks like this patch set causes conflicts right now. Can you rebase and force push a single commit with your changes?

SDL_ttf.h
```diff
@@ -426,6 +413,32 @@ extern DECLSPEC SDL_bool TTF_GetFontSDF(const TTF_Font *font);
...
+   This functions returns always 0, or -1 if SDL_ttf is not compiled with HarfBuzz
+
+   Deprecated. Prefer TTF_SetFontDirection() and TTF_SetFontScriptName()
+*/
+extern DECLSPEC int SDLCALL TTF_SetDirection(int direction); /* hb_direction_t */
+extern DECLSPEC int SDLCALL TTF_SetScript(int script); /* hb_script_t */
+
+/* Set direction and script per font 
+   'script' is null terminated string of exactly 4 characters.
+   This functions returns 0, or -1 if SDL_ttf is not compiled with HarfBuzz or invalid parameter
```
sezero, line 437: `These functions return 0, or -1 if SDL_ttf is not compiled with HarfBuzz or in case of an invalid parameter.`

1bsyl:
@slouken Not sure but it says "This branch has no conflicts with the base branch"

SDL_ttf.h
```diff
@@ -425,16 +425,16 @@ typedef enum
 } TTF_Direction;
 
 /* Set Direction and Script to be used for text shaping.
-   This functions returns always 0, or -1 if SDL_ttf is not compiled with HarfBuzz
+   These functions returns 0, or -1 if SDL_ttf is not compiled with HarfBuzz
```
sezero, line 428: `s/returns/return/` :)

sezero:
> @slouken Not sure but it says "This branch has no conflicts with the base branch" when selecting "squash and merge"

But not when wants to `rebase and merge`:


That usually seems to happen when one merges master into the branch once or multiple times

slouken:
Merged!

PR #180, opened 2022-02-20
Small Bugfix: Possible Buffer Overflow

manuel-fischer:
If the allocation fails, the next time the function is called with the same string, it seems that the buffer size is enough. A buffer overflow occurs.


1bsyl:
Thanks fixed

PR #181, opened 2022-02-22
Sized String Support = Optimizations

manuel-fischer:
Added sized string support, such that it can be used for example with C++ `std::string_view`. Introducing an internal `StringView` struct, that contains a pointer, the size of the string and the encoding.

This allows to reduce string allocations, which were needed to convert to UTF8 and for splitting the strings in wrapped rendering. Instead I have solved both problems by almost always using sized strings and by moving the encoding specific string operations deeper into the call chain.

Also the way how layouting with the position buffer works is changed. The arrays are extracted from the font struct into an own position array, the struct contains a dummy element for ABI stability. The new array-struct is passed as an explicit parameter, which can be left `NULL`, to avoid allocation in `TTF_Size_Internal`.

## New functions
Each function related to rendering text has added an additional parameter `size_t text_size` after the `text` parameter.
- `TTF_Size*_SZ`
- `TTF_Measure*_SZ`
- `TTF_Render*_Solid_SZ`
- `TTF_Render*_Shaded_SZ`
- `TTF_Render*_Blended_SZ`
- `TTF_Render*_Solid_Wrapped_SZ`
- `TTF_Render*_Shaded_Wrapped_SZ`
- `TTF_Render*_Blended_Wrapped_SZ`

`*` represents all of `TEXT`, `UTF8`, `UNICODE`

## For the documentation of the new functions
For UTF8 and TEXT `text_size` denotes the number of bytes, for UNICODE it denotes the number of words (number of bytes / 2)

`NULL` can be passed as a paramter, as long as the size is also `0`.
The range `[text, text+text_size)` should be a valid string and thus a valid memory range.

## Showfont
Added new command line parameter `[-wrap width]` which forces to use the `*_Wrapped` functions with `width` as the `wrapLength` parameter.

## Problem
One problem, that might need to be discussed is the combinatoric explosion of rendering functions for the different formats.

### DIFF
--- a/SDL_ttf.c
+++ b/SDL_ttf.c
@@ -117,7 +117,7 @@ int TTF_SetScript(int script) /* hb_script_t */
 #define HAVE_BLIT_GLYPH_32
 
 /* Use Duff's device to unroll loops */
-//#define USE_DUFFS_LOOP
+/*#define USE_DUFFS_LOOP*/
 
 #if defined(HAVE_SSE2_INTRINSICS)
 static SDL_INLINE int hasSSE2()
@@ -581,11 +581,11 @@ static SDL_INLINE void BG_Blended_Opaque_SSE(const TTF_Image *image, Uint32 *des
         /* *INDENT-OFF* */
         DUFFS_LOOP4(
             /* Read 16 Uint8 at once and put into 4 __m128i */
-            s  = _mm_loadu_si128(src);          // load unaligned
-            d0 = _mm_load_si128(dst);           // load
-            d1 = _mm_load_si128(dst + 1);       // load
-            d2 = _mm_load_si128(dst + 2);       // load
-            d3 = _mm_load_si128(dst + 3);       // load
+            s  = _mm_loadu_si128(src);          /* load unaligned */
+            d0 = _mm_load_si128(dst);           /* load */
+            d1 = _mm_load_si128(dst + 1);       /* load */
+            d2 = _mm_load_si128(dst + 2);       /* load */
+            d3 = _mm_load_si128(dst + 3);       /* load */
 
             L  = _mm_unpacklo_epi8(zero, s);
             H  = _mm_unpackhi_epi8(zero, s);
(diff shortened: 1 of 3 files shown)


sezero:
The patch does not apply: you generated against an older git revision. It seems like you miss this commit: https://github.com/libsdl-org/SDL_ttf/commit/8368ea92c8bb86d65dd28ba10eeeeecefbd7d214

SDL_ttf.c
```diff
@@ -3329,82 +3516,167 @@ SDL_Surface* TTF_RenderGlyph32_Shaded(TTF_Font *font, Uint32 ch, SDL_Color fg, S
...
+            }
+            new_lines = (TTF_StringView *)SDL_realloc(*p_lines, maxNumLines * sizeof (*p_lines));
+            if (new_lines == NULL) {
+                SDL_OutOfMemory();
+                return SDL_FALSE;
+            }
+            *p_lines = new_lines;
+        }
+
+        TTF_StringView *line = &(*p_lines)[*p_numLines];
```
sezero, line 3598: This will hit `-Werror=declaration-after-statement`

manuel-fischer:
> The patch does not apply: you generated against an older git revision. It seems like you miss this commit: [8368ea9](https://github.com/libsdl-org/SDL_ttf/commit/8368ea92c8bb86d65dd28ba10eeeeecefbd7d214)

It was a bug, that I have found while rewriting the code, it can be discarded, because I have rewritten that part in the code.

showfont.c
```diff
@@ -266,36 +277,70 @@ int main(int argc, char *argv[])
...
     /* Render and center the message */
-    if (argc > 2) {
-        message = argv[2];
+    if (argc >= 2) {
+        if(wrap) // concatenate message
+        {
+            size_t start = 0;
+            size_t rest_bytes = sizeof(string);
+            
+            for(int i = 2; i < argc && rest_bytes > 0; ++i)
```
sezero, line 286: This will fail build in C90 mode.

sezero:
>  it can be discarded, because I have rewritten that part in the code.

You should rebase your patchset, instead.

SDL_ttf.c
```diff
@@ -2985,13 +3101,10 @@ static int TTF_Size_Internal(TTF_Font *font,
         int y_offset  = hb_glyph_position[g].y_offset;
 #else
     /* Load each character and sum it's bounding box */
-    textlen = SDL_strlen(text);
```
sezero, line 2988: The local `size_t textlen;` should be removed from the top of the procedure, as well.

1bsyl:
I also want to look at this (end of next week probably).
Quickly:
- there are still things commented out you could remove (About pa allocation).
- do we really need a posbuf array.(is this reallocated more than once, since we use a new one each time). But I may need to read more precisely

manuel-fischer:
>     * do we really need a posbuf array.(is this reallocated more than once, since we use a new one each time). But I may need to read more precisely

Multiple Reallocations do happen for the posbuf, while previously the posbuf is permanently saved in the font and not resized, if it was large enough. But this could be improved by either allocating directly by the amount of glyphs in the string or by rewriting it to not use allocated arrays at and iterate again through the string characters. Note that for the latter, the glyph-lookup would not change, because it is already looked up again, when iterating through the posbuf-array.

1bsyl:
ok, I've looked and tried (but not fully)
Just to make sure I saw no regression with some multi lang test I have.
I also run a fuzzy test, and it didn't crash.

Maybe not a problem, but there is lot of modification in the wrap version. and there were corner cases and special behavior (with wrap-length == 0 for instance). hope we have no regression.

I see now it decodes  unicode and text, instead of converting to utf8, which somehow requires more testing in case of bug. So not sure if this is better.
As you said, API is longer with all _SZ variants for string_view.

Measuring the time with some bench, I didn't see a noticeable speed improvement,
so I wonder if this worth adding more complexity in sdl_ttf code
and if string_view API could just be built on top of RenderUTF8 function.

slouken:
SDL_ttf has been updated to remove Latin-1 and UNICODE API surfaces and support strings with length by default.

PR #190, opened 2022-03-31
Define DLL_EXPORT when building shared libraries for Windows

cgutman:
Right now, a CMake build with `-DBUILD_SHARED_LIBS=ON` ends up building a DLL with no exports (or lib file). `DLL_EXPORT` must be defined for our exports to actually be exported.


slouken:
Does this change need to be made for SDL_image and SDL_mixer as well?

cgutman:
It looks like they both already handle it.

https://github.com/libsdl-org/SDL_image/blob/ce7b232f58e0a7f0dcaa03c68fdaebefae5c8167/CMakeLists.txt#L102-L104

https://github.com/libsdl-org/SDL_mixer/blob/9425f0e422e1b2dea3d8e9a43f69c2209ba8e2bc/CMakeLists.txt#L91-L93

slouken:
Okay, great.

PR #200, opened 2022-05-05
Use minor version for feature releases

smcv:
Similar to https://github.com/libsdl-org/SDL_mixer/pull/383. This is my last PR like this, unless someone is particularly interested in SDL_rtf (which isn't in the Steam Runtime or Debian).

* Revert "Make it possible to build with older SDL (for GitHub CI)"
    
    SDL < 2.0.10 would have freed the memory with SDL_free(pixels)
    rather than SDL_SIMDFree(pixels), which is invalid when pixels is a
    pointer into the middle of a malloc-allocated block.
    
    SDL_SIMDFree(pixels) expects the pointer `pixels` to be like this,
    with a poitner to the malloc'd block "behind" the texture data:
    
                  sizeof(void*)
                    |<--->|
        [ unused    | ptr | texture data...               ]
           ^                 ^
           |                 |
          ptr               pixels
    
        |<-- malloc'd block                            -->|
    
    which is also what this function has been constructed to do.
    
    This reverts commit 9a86e3f1e167f9d117a1723c67c802cda1a48e3a.

* Relax required SDL version to 2.0.10
    
    Commit 99f64c8 bumped the dependency to 2.0.12, but in fact 2.0.10
    should be enough for SDL_SIMD_ALIGNED.

* workflows: Add basic CI using Github Actions

* workflows: Run showfont
    
    We can't be very thorough here, but this is a simple check that it's
    working at all.

* build: Use minor version number for feature releases
    
    Similar to what has been done in SDL itself, this allows micro versions
    with an odd minor version to be used as prereleases for a stable
    release. It also allows micro versions with an even minor version to
    be used to fix regressions in their corresponding .0 release
    if that becomes necessary.

* cmake: On Unix, make the library ABI-compatible with the Autotools build
    
    This is not yet implemented for macOS, only for generic Unix (in practice
    meaning Linux and *BSD).


smcv:
It's going to be a bit awkward to get CI going for this one, since it needs a newer SDL than the version in Ubuntu 20.04. We'll have to either build SDL from source, or use a PPA, or use a Debian 11 or Ubuntu 22.04 container.

sezero:
It requires 2.0.12+, isn't that available on CI hosts?

smcv:
Ubuntu 20.04 only had 2.0.10, unfortunately.

slouken:
We can fix it so it works with SDL 2.0.10, I'll make that change.

slouken:
This should work: https://github.com/libsdl-org/SDL_ttf/commit/9a86e3f1e167f9d117a1723c67c802cda1a48e3a

smcv:
Isn't that going to make the older SDL allocate unaligned memory, which will cause a fault with SIMD instructions?

slouken:
Yes, but there's no help for it. I think the older SDL used less aggressive SIMD optimizations and for those the default system allocator alignment was fine. We can add a #warning if you want, but the goal is to enable CI, not necessarily solve running with really old SDL, so I think this is fine.

sezero:
OK, the project still needs setting up for CI. (How exactly is it done btw?)

smcv:
The CMake build is still failing, I'll come back to this tomorrow.

slouken:
> OK, the project still needs setting up for CI. (How exactly is it done btw?)

All the SDL repositories are already setup for GitHub Actions. No additional work is needed.

sezero:
> This should work: [9a86e3f](https://github.com/libsdl-org/SDL_ttf/commit/9a86e3f1e167f9d117a1723c67c802cda1a48e3a)

Would it not be better to define this to the actual value 0x08,
i.e. one may build against an old SDL but run against a new one?
Or, would it really misbehave when running against an old SDL?

smcv:
I think 9a86e3f might make SDL free the texture memory incorrectly, both with old and new SDL, but I haven't fully worked through the logic of it yet. It seems a risky change to be making, and it might be better to make the CI either compile its own up-to-date SDL, or run in a newer container.

sezero:
> I think [9a86e3f](https://github.com/libsdl-org/SDL_ttf/commit/9a86e3f1e167f9d117a1723c67c802cda1a48e3a) might make SDL free the texture memory incorrectly if compiled with an old SDL, with either an old or new SDL at runtime

That would be scary, yes.

smcv:
> Otherwise, we can revert https://github.com/libsdl-org/SDL_ttf/commit/9a86e3f1e167f9d117a1723c67c802cda1a48e3a and change minimum required SDL2 ver from 2.0.12 to 2.0.10

That sounds like a good approach to me.

sezero:
OK, this is in

Issue #187, opened 2022-03-19
Arbitrary memory overwrite occurs when loading glyphs and rendering text with a malformed TTF file.


ch4rli3kop:
Hello, I found a vulnerability in this project.

### Summary

Arbitrary memory overwrite occurs when loading glyphs and rendering text with a malformed TTF file.

### System Info

- Operating System: Ubuntu 20.04

### Detailed Description

When the function `TTF_RenderText_Solid()` is executed, it internally calls `TTF_Size_Internal()` and `Render_Line()`. Since the code load and render glyph data after measuring bitmap size, if the measured size has a problem, it causes memory overflow or arbitrary memory write when rendering the data. The bitmap size of glyph affects variables `xstart` and `ystart`. And they are used to calculate the destination of `BG_SSE()`. Therefore a malformed TTF file that has manipulated glyph data will result in memory corruption.

If the rendered string has only characters that mapped malformed glyph data, `ft_failure` occurs when calling `FT_Render_Glyph`. But, if the string has a character that mapped normally formed glyph data in front of the mal-mapped character, the normal character is rendered with corrupted size while `FT_Render_Glyph` is normally called. So, the normal character's glyph data is overwritten to arbitrary memory addresses with corrupted `xstart` and `ystart`. The address will be heap or stack.

In the below code and attached malformed TTF file, a character "T" has normal glyph data and a character "S" has malformed glyph data. Since The address of the calculated destination with `xstart` and `ystart` is not a valid memory address, a segmentation fault occurs. debugged data is below.

```shell
Starting program: /home/ch4rli3kop/SDL/fuzz/OpenTTF crashed/access_violation_0000xxxxxxxxx9E4_0000xxxxxxxxx800_1
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
INFO: Found charmap: platform id 0, encoding id 280

Program received signal SIGSEGV, Segmentation fault.
0x00007ffff7c25318 in _mm_load_si128 (__P=0x7fffd9fc71a0) at /usr/lib/gcc/x86_64-linux-gnu/9/include/emmintrin.h:697
697	  return *__P;

[ Legend: Modified register | Code | Heap | Stack | String ]
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────── registers ────
$rax   : 0x7fffd9fc71a0    
[...]

PR #204, opened 2022-05-09
Check for overflow more carefully

smcv:
* Factor out repeated code to allocate a SDL_Surface

* Check for integer overflow when allocating surfaces
    
    On practical computers, doing memory-size calculations in the signed
    integer domain should work fine; but it could be technically possible
    for a crafted TTF file to make the height or width negative, resulting
    in signed integer overflow, which is formally undefined behaviour
    (although gcc -fwrapv defines it to wrap around as twos-complement).
    
    To make it more obvious that this is right, do all size calculations
    in AllocateAlignedPixels in the unsigned size_t domain, with
    range-checks that do not rely on the behaviour at overflow.

---

Works in trivial testing with `showfont`, but I haven't tried this with the malformed font from #187 yet.

### DIFF
--- a/SDL_ttf.c
+++ b/SDL_ttf.c
@@ -1366,19 +1366,27 @@ static SDL_INLINE int Render_Line(const render_mode_t render_mode, int subpixel,
 #endif
 }
 
-static SDL_Surface* Create_Surface_Solid(int width, int height, SDL_Color fg, Uint32 *color)
+
+/* Create a surface with memory:
+ * - pitch is rounded to alignment
+ * - address is aligned
+ *
+ * If format is 4 bytes per pixel, bgcolor is used to initialize each
+ * 4-byte word in the image data.
+ *
+ * Otherwise, the low byte of format is used to initialize each byte
+ * in the image data.
+ */
+static SDL_Surface *AllocateAlignedPixels(int width, int height, SDL_PixelFormatEnum format, Uint32 bgcolor)
 {
     const int alignment = Get_Alignement() - 1;
-    SDL_Surface *textbuf;
+    const int bytes_per_pixel = SDL_BYTESPERPIXEL(format);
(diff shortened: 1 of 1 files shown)


smcv:
I haven't checked all the allocations in `SDL_ttf.c` for attacker-controllability, only the ones using the same code as in #187. Ideally all the size arithmetic would work like this?

The overflow-checked arithmetic functions are basically libsdl-org/SDL#5643, but I inlined them here to avoid depending on a version of SDL that doesn't exist yet. After they're in SDL, ideally the version here would use `SDL_VERSION_ATLEAST` to try to use SDL's copy, falling back to a local copy if SDL is too old.

SDL_ttf.c
```diff
@@ -1366,28 +1366,81 @@ static SDL_INLINE int Render_Line(const render_mode_t render_mode, int subpixel,
...
+
+/* TODO: Use SDL_size_mul_overflow after it's in a release */
+SDL_FORCE_INLINE int size_mul_overflow (size_t a,
+                                        size_t b,
+                                        size_t *ret)
 {
-    const int alignment = Get_Alignement() - 1;
-    SDL_Surface *textbuf;
-    Sint64 size;
+#if defined(_SDL_HAS_BUILTIN) && _SDL_HAS_BUILTIN(__builtin_mul_overflow)
```
smcv, line 1378: Looks like this doesn't do what I wanted it to if `_SDL_HAS_BUILTIN` isn't defined, so it'll need more backwards-compat goo.

slouken: How about just do the C versions of those functions and don't worry about the builtins, and use the SDL versions if built with a newer SDL?

smcv:
> After they're in SDL, ideally the version here would use SDL_VERSION_ATLEAST to try to use SDL's copy, falling back to a local copy if SDL is too old

Done.

PR #210, opened 2022-05-11
Add support for vendored libraries in CMake

### LINKED ISSUE #101: [PATCH] CMake: small error and code cleanup
# This bug report was migrated from our old Bugzilla tracker.

These attachments are available in the static archive:

* [CMake.patch (CMake.patch, text/plain, 2018-11-14 08:28:19 +0000, 2297 bytes)](https://bugzilla.libsdl.org/attachment.cgi?id=3488)

**Reported in version:** unspecified
**Reported for operating system, platform:** Windows 10, x86

# Comments on the original bug report:

On 2018-11-14 08:28:19 +0000, David Demelier wrote:

> Created attachment 3488
> CMake.patch
> 
> Hello,
> 
> I'm sorry the initial CMake support got a small error that made it non-working on Linux system because I only have tested it on a case-insensitive OS.
> 
> The patch changes the file names to the lowercase convention to avoid this kind of errors.
> 
> sdl2_ttf-config.cmake
> sdl2_ttf-config-version.cmake
> sdl2_ttf-config-targets.cmake
> 
> I propose to do the same for SDL and will do it if your okay (it's fully backward compatible).

On 2018-11-14 08:29:17 +0000, David Demelier wrote:

> To use SDL2_ttf on CMake project:
> 
>     find_package(SDL2_ttf)
>     target_link_libraries(myexe SDL2::SDL2_ttf)

### LINKED ISSUE #208: Build SDL_ttf with vendored libraries
I suggest add the ability to build SDL_ttf with vendored libraries (freetype and harfbuzz) the same way as SDL_image are builded with libpng/libjpeg/etc.

For example in SDL_image

```
        if (SUPPORT_JPG_VENDORED)
            set(jpeg OFF)
            add_subdirectory(external/jpeg EXCLUDE_FROM_ALL)
            add_library(JPEG::JPEG ALIAS jpeg)
            list(APPEND INSTALL_EXTRA_TARGETS jpeg)
        else()
            find_package(JPEG REQUIRED)
            list(APPEND PC_REQUIRES libjpeg)
        endif()
```

madebr:
- Adds `TTF_WITH_HARFBUZZ` option to enable/disable harfbuzz
- Adds `TTF_WITH_FREETYPE_VENDORED` and `TTF_WITH_HARFBUZZ_VENDORED` options to enable/disable vendoring.
- The export of targets + generation of the .pc file was already there, so I only needed to add support for the vendored libraries there
- add 'd' suffix for debug libraries
- allow parallel installation of shared/static libraries in the same install prefix (mostly useful for linux distributions)

A prefix might look like this:
```
prefix
├── include
│   └── SDL2
│       └── SDL_ttf.h
├── lib64
│   ├── cmake
│   │   └── SDL2_ttf
│   │       ├── SDL2_ttfConfig.cmake
│   │       ├── SDL2_ttfConfigVersion.cmake
│   │       ├── SDL2_ttf-shared-targets.cmake
│   │       ├── SDL2_ttf-shared-targets-debug.cmake
│   │       ├── SDL2_ttf-shared-targets-release.cmake
│   │       ├── SDL2_ttf-static-targets.cmake
│   │       ├── SDL2_ttf-static-targets-debug.cmake
│   │       └── SDL2_ttf-static-targets-release.cmake
│   ├── libSDL2_ttf-2.0d.so -> libSDL2_ttf-2.0d.so.0
│   ├── libSDL2_ttf-2.0d.so.0 -> libSDL2_ttf-2.0d.so.0.1900.0
│   ├── libSDL2_ttf-2.0d.so.0.1900.0
│   ├── libSDL2_ttf-2.0.so -> libSDL2_ttf-2.0.so.0
│   ├── libSDL2_ttf-2.0.so.0 -> libSDL2_ttf-2.0.so.0.1900.0
│   ├── libSDL2_ttf-2.0.so.0.1900.0
│   ├── libSDL2_ttf.a
│   ├── libSDL2_ttfd.a
│   ├── libSDL2_ttfd.so -> libSDL2_ttf-2.0d.so.0
│   ├── libSDL2_ttf.so -> libSDL2_ttf-2.0.so.0
│   └── pkgconfig
│       └── SDL2_ttf.pc
└── share
    └── licenses
        └── SDL2_ttf
            └── COPYING.txt
```

@rmg-nik can you please test this?

Fixes #101
Fixes #208


sezero:
There are multiple cmake warnings from the CI run:
https://github.com/libsdl-org/SDL_ttf/runs/6397113845?check_suite_focus=true

This one isn't from us, but from harfbuzz (do we want to fix that??)
```
CMake Warning (dev) at external/harfbuzz/CMakeLists.txt:49 (option):
  Policy CMP0077 is not set: option() honors normal variables.  Run "cmake
  --help-policy CMP0077" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.

  For compatibility with older versions of CMake, option is clearing the
  normal variable 'HB_BUILD_UTILS'.
This warning is for project developers.  Use -Wno-dev to suppress it.
```

The following are from our cmake'ry:
```
CMake Warning (dev) at /usr/local/share/cmake-3.23/Modules/CheckIncludeFile.cmake:82 (message):
  Policy CMP0075 is not set: Include file check macros honor
  CMAKE_REQUIRED_LIBRARIES.  Run "cmake --help-policy CMP0075" for policy
  details.  Use the cmake_policy command to set the policy and suppress this
  warning.

  CMAKE_REQUIRED_LIBRARIES is set to:

    m

  For compatibility with CMake 3.11 and below this check is ignoring it.
Call Stack (most recent call first):
  external/harfbuzz/CMakeLists.txt:92 (check_include_file)
This warning is for project developers.  Use -Wno-dev to suppress it.

[...]

-- Configuring done
CMake Warning at CMakeLists.txt:95 (add_library):
  Cannot generate a safe runtime search path for target SDL2_ttf because
  files in some directories may conflict with libraries in implicit
  directories:

    runtime library [libfreetype.so.6] in /usr/lib/x86_64-linux-gnu may be hidden by files in:
      /home/runner/work/SDL_ttf/SDL_ttf/build/external/freetype

  Some of these libraries may not be found correctly.


CMake Warning at CMakeLists.txt:266 (add_executable):
  Cannot generate a safe runtime search path for target glfont because files
  in some directories may conflict with libraries in implicit directories:

    runtime library [libfreetype.so.6] in /usr/lib/x86_64-linux-gnu may be hidden by files in:
      /home/runner/work/SDL_ttf/SDL_ttf/build/external/freetype

  Some of these libraries may not be found correctly.


CMake Warning at CMakeLists.txt:266 (add_executable):
  Cannot generate a safe dependent library path for target glfont because
  files in some directories may conflict with libraries in implicit
  directories:

    runtime library [libfreetype.so.6] in /usr/lib/x86_64-linux-gnu may be hidden by files in:
      /home/runner/work/SDL_ttf/SDL_ttf/build/external/freetype

  Some of these libraries may not be found correctly.


CMake Warning at CMakeLists.txt:267 (add_executable):
  Cannot generate a safe runtime search path for target showfont because
  files in some directories may conflict with libraries in implicit
  directories:

    runtime library [libfreetype.so.6] in /usr/lib/x86_64-linux-gnu may be hidden by files in:
      /home/runner/work/SDL_ttf/SDL_ttf/build/external/freetype

  Some of these libraries may not be found correctly.


CMake Warning at CMakeLists.txt:267 (add_executable):
  Cannot generate a safe dependent library path for target showfont because
  files in some directories may conflict with libraries in implicit
  directories:

    runtime library [libfreetype.so.6] in /usr/lib/x86_64-linux-gnu may be hidden by files in:
      /home/runner/work/SDL_ttf/SDL_ttf/build/external/freetype

  Some of these libraries may not be found correctly.
```

madebr:
I"ve seen those warnings about the runtime path too.
From what I understand, that cmake detects 2 freetype shared libraries in the search path: the system libfreetype & the vendored libfreetype.
It warns that running the executables might use the wrong library (you should use `LD_LIBRARY_PATH` to select the correct one).
I'm not sure how to solve this.

madebr:
@sezero 
Can you please check the following pr's + bump the git submodules of this repo?
- https://github.com/libsdl-org/freetype/pull/1
- https://github.com/libsdl-org/harfbuzz/pull/1

slouken:
This is failing CI:
https://github.com/libsdl-org/SDL_ttf/runs/6398693946?check_suite_focus=true
```
CMake Error at build/cmake_install.cmake:190 (file):
  file INSTALL cannot find "//SDL2_ttf.pc": No such file or directory.
```

slouken:
Also, we shouldn't be installing freetype or harfbuzz. If we build with the versions included with SDL_ttf they should be linked statically and the symbols should not be exported.

madebr:
If SDL2_{image,ttf} is built as a shared library, the vendored libraries will also be built as shared libraries.
Vice versa for static libraries.
So currently, a shared SDL2_ttf will never have libfreetype/libharfbuzz built-in.

Instead, do you want to build one single SDL2_ttf.(so|dll|dylib) library?

slouken:
> Instead, do you want to build one single SDL2_ttf.(so|dll|dylib) library? =always build a static freetype/libharfbuzz

Yes, this is how we build SDL_ttf with other build systems.

This isn't true for SDL_image and SDL_mixer, but they should use already installed dependency libraries if they are available.

madebr:
> > Instead, do you want to build one single SDL2_ttf.(so|dll|dylib) library? =always build a static freetype/libharfbuzz
> 
> Yes, this is how we build SDL_ttf with other build systems.

Ok then, that willl be in the next batch of changes.

> This isn't true for SDL_image and SDL_mixer, but they should use already installed dependency libraries if they are available.

It is also possible with SDL_image when dynamic loading is disabled.

sezero:
> @sezero Can you please check the following pr's + bump the git submodules of this repo?
> 
>     * [Make sure freetype uses the same harfbuzz library as SDL2_ttf freetype#1](https://github.com/libsdl-org/freetype/pull/1)
> 
>     * [Use freetype target instead of FindFreetype target to avoid finding system libraries harfbuzz#1](https://github.com/libsdl-org/harfbuzz/pull/1)

AFAICS you already did this - is there still work for me to do?

sezero:
In CI cmake build log, I see freetype objects being built but not harfbuzz:
something wrong with me or why is that happening?

rmg-nik:
@madebr Hi
I tested it via the cmake project https://github.com/rmg-nik/sdl_sandbox where all SDL libraries are submodules.
The main goal is creating simple cmake project "sdl sanbox", where with minimum efforts anybody can work with SDL (no install libraries, no separate compilation, no package managers hell, no dependencies).
So all libraries should be static to achieve the goal (flag option(BUILD_SHARED_LIBS "Build the library as a shared library" OFF) in root CMakeLists.txt https://github.com/rmg-nik/sdl_sandbox/blob/main/CMakeLists.txt)
There is a some problems.
First, I need fix SDL_image/SDL_mixer/SDL_ttf CMakeLists.txt
SDL_image:
```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 225526e..ad5b73e 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,7 +55,7 @@ project(SDL2_image LANGUAGES C

 message(STATUS "Configuring ${PROJECT_NAME} ${PROJECT_VERSION}")

-if (NOT ANDROID AND NOT (TARGET SDL2::SDL2 OR TARGET SDL2::SDL2-static))
+if (NOT (TARGET SDL2::SDL2 OR TARGET SDL2::SDL2-static))
     find_package(SDL2 REQUIRED)
 endif()

@@ -205,7 +205,7 @@ endfunction()

 # Workaround for Ubuntu 20.04's SDL being older than
 # https://github.com/libsdl-org/SDL/issues/3531
-if (NOT TARGET SDL2::SDL2)
+if (NOT (TARGET SDL2::SDL2 OR TARGET SDL2::SDL2-static))
     find_library(SDL2_LIBRARY
         NAMES SDL2
         HINTS "${SDL2_EXEC_PREFIX}"
```

SDL_mixer:
```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index af88a92d..47e74159 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,7 +25,7 @@ option(SUPPORT_MID_TIMIDITY "Support loading MIDI music via TiMidity" ON)

 option(BUILD_SHARED_LIBS "Enable shared library" ON)

-if (NOT ANDROID AND NOT (TARGET SDL2 OR TARGET SDL2-static))
+if (NOT (TARGET SDL2::SDL2 OR TARGET SDL2::SDL2-static))
     find_package(SDL2 REQUIRED)
     if(NOT TARGET SDL2::SDL2)
         # SDL < 2.0.12
@@ -166,4 +166,9 @@ if(WIN32 AND BUILD_SHARED_LIBS)
 endif()

 target_include_directories(SDL2_mixer PUBLIC include)
-target_link_libraries(SDL2_mixer PRIVATE SDL2::SDL2)
+
+if (NOT BUILD_SHARED_LIBS AND TARGET SDL2::SDL2-static)
+  target_link_libraries(SDL2_mixer PRIVATE SDL2::SDL2-static)
+else()
+  target_link_libraries(SDL2_mixer PRIVATE SDL2::SDL2)
+endif()
```

SDL_ttf:
```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 805316b..d6d4b1e 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -62,7 +62,7 @@ endif()

 # Workaround for Ubuntu 20.04's SDL being older than
 # https://github.com/libsdl-org/SDL/issues/3531
-if (NOT TARGET SDL2::SDL2)
+if (NOT (TARGET SDL2::SDL2 OR TARGET SDL2::SDL2-static))
     find_library(SDL2_LIBRARY
         NAMES SDL2
         HINTS "${SDL2_EXEC_PREFIX}"
```

After that I can configure and compile project, but the compiler cannot find SDL_ttf.h file, when with SDL_mixer.h/SDL_image.h all is right.

So, first - can we modify SDL_image/SDL_mixer with diff above? And second, how properly add SDL_ttf.h to SDL_ttf target so dependent project can access it?

madebr:
> In CI cmake build log, I see freetype objects being built but not harfbuzz: something wrong with me or why is that happening?

https://github.com/libsdl-org/SDL_ttf/runs/6404634378?check_suite_focus=true#step:7:52 + line 53
This contains libharfbuzz.so.
But it looks like libfreetype is not linking to harfbuzz, so I might need to look at the freetype sources again.

sezero:
> https://github.com/libsdl-org/SDL_ttf/runs/6404634378?check_suite_focus=true#step:7:52 + line 53 This contains libharfbuzz.so.

What I mean is that I don't see harfbuzz itself being compiled

> But it looks like libfreetype is not linking to harfbuzz, so I might need to look at the freetype sources again.

IIRC, freetype and harfbuzz has a chicken-egg problem, circularly depending on each other
and the autotools side worked around that by some foul magic.

madebr:
> > https://github.com/libsdl-org/SDL_ttf/runs/6404634378?check_suite_focus=true#step:7:52 + line 53 This contains libharfbuzz.so.
> 
> What I mean is that I don't see harfbuzz itself being compiled

Line 52 contains: `-o external/harfbuzz/libharfbuzz.so`
Line 53 uses external/harfbuzz/libharfbuzz.so in SDL2_ttf.so

> > But it looks like libfreetype is not linking to harfbuzz, so I might need to look at the freetype sources again.
> 
> IIRC, freetype and harfbuzz has a chicken-egg problem, circularly depending on each other and the autotools side worked around that by some foul magic.

I saw that too yesterday, I'll need to investigate how CMake handles this.

madebr:
> So it's using some cached result??

We haven't configured any caching on github CI, afaik.
harfbuzz consists of a single source `harfbuzz.cc`, which is compiled at line 51 and linked into a shared library at line 52.
Line 53 uses the shared library.

I hope that answers the question of building harfbuzz?

sezero:
>  harfbuzz consists of a single source `harfbuzz.cc`, 

Huh.. Missed that, sorry for the noise.

sezero:
> So the solution is to always build both of them as a static library. Does autotools the same thing?

Practically yes it does, and I agree with this solution

slouken:
> ```
> CMake Error: The inter-target dependency graph contains the following strongly connected component (cycle):
>   "harfbuzz" of type SHARED_LIBRARY
>     depends on "freetype" (weak)
>   "freetype" of type SHARED_LIBRARY
>     depends on "harfbuzz" (weak)
> At least one of these targets is not a STATIC_LIBRARY.  Cyclic dependencies are allowed only among static libraries.
> CMake Generate step failed.  Build files cannot be regenerated correctly.
> ```
> 
> So the solution is to always build both of them as a static library. Does autotools the same thing?

Autotools goes so far as to build them both shared when using the system libraries, and actually include the source directly when building using the included libraries:
https://github.com/libsdl-org/SDL_ttf/blob/main/Makefile.am

We don't need to do that with CMake, but that's how autotools solves it.

PR #212, opened 2022-05-12
Fix workaround causing issues on static builds

object71:


sezero:
@madebr ?

madebr:
@object71 

Hi!
I've just created https://github.com/libsdl-org/SDL_ttf/pull/213 that solves this in a very similar but slightly different way.
Can you please test that one and check whether your static build error goes away?

object71:
> @object71
> 
> Hi! I've just created #213 that solves this in a very similar but slightly different way. Can you please test that one and check whether your static build error goes away?

Works. I had a similar issue with SDL_image too

madebr:
Thanks for checking!
I'm currently reworking the cmake build systems of SDL_image, SDL_mixer and SDL_ttf.
In the end, all should follow the same pattern and behave in the same way.

object71:
@madebr Okay, closing my pull requests then

PR #213, opened 2022-05-12
Cmake work

### LINKED ISSUE #141: SDL2_ttfConfig.cmake is missing from release archives
**Version:** 2.0.15

I'm not sure if there's something obvious I'm missing, but as the title says I couldn't find `SDL2_ttfConfig.cmake` in any of the tarballs on [libsdl.org](https://www.libsdl.org/projects/SDL_ttf/release/). However, the file exists in the release found on GitHub. This is a problem because when installing the library through my Linux distro you can't use `find_package(SDL_ttf)` to link against it through CMake without some sort of clumsy workaround. This may not be the correct place to report this, but I figured I would start here.

I also noticed that the tarball from GitHub contains `.hgignore`, not sure if this is intentional.

### LINKED ISSUE #143: Better Support for projects that uses add_subdirectory
the current **cmake** file assumes the user is going to install it, however I usually prefer to build standalone projects, which is why I add most external libs as sub directories via **add_subdirectory**

the current **cmake** file, requires the consumer of lib to manually copy **SDL_ttf.h** to his/her **src** directory

however, that can be easily solved by just adding this line to SDL_ttf cmake file
`target_include_directories(SDL2_ttf PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>)`

then any consumer of the lib can simply add via `add_subdirectory` and immediately include the header and link it.

### LINKED ISSUE #226: CMake package config tries to include non-existent SDL_ttf-targets.cmake file
The cmake package config file `SDL2_ttfConfig.cmake` tries to `include` the file `SDL_ttf-targets.cmake` ([at the bottom of that file](https://github.com/libsdl-org/SDL_ttf/blob/main/SDL2_ttfConfig.cmake.in#L18)). But, the file generated by the build is called:  `SDL2_ttf-${sdl2_ttf_install_name_infix}-targets.cmake` (https://github.com/libsdl-org/SDL_ttf/blob/main/CMakeLists.txt#L276)

As a result, the `include` fails with an error similar to the following:

```
CMake Error at <SDL_ttf install prefix>/cmake/SDL2_ttfConfig.cmake:18 (include):
  include could not find requested file:

    <SDL_ttf install prefix>/cmake/SDL2_ttf-targets.cmake
Call Stack (most recent call first):
  CMakeLists.txt:276 (find_package)
 ```
 
A possible solution is to change the name of the target file inside `SDL_ttfConfig.cmake` like so:

    include("${CMAKE_CURRENT_LIST_DIR}/SDL2_ttf-@sdl2_ttf_install_name_infix@-targets.cmake")
    
The [`configure_package_config_file`](https://github.com/libsdl-org/SDL_ttf/blob/main/CMakeLists.txt#L267) call should then expand this to the right target file name.

madebr:
- Use `SDL2TTF_` prefix for all options
- Create SDL2_ttf::SDL2_ttf when building a shared library,
  create SDL2_ttf::SDL2_ttf-static when building a static library
- Use same CMake layout as SDL_image/SDL_mixer
- Create libSDL2_ttfd.so symbolic link (when building debug build type)
- Add PrivateSdlFunctions.cmake script for common functionality between SDL_image/SDL_mixer
- Add FindXXX.cmake scripts for dependencies that don't provide cmake configuration files 
- Add CMakeLists.txt + cmake scripts to source distribution (EXTRA_DIST of autotools)
- Test CMake on mingw64 (using system dependencies there)
- Add MacOS CI
- Create symbolic link at build time + install it
- Only install SDL2_ttf.pc when building a shared SDL2_ttf (to avoid overwriting when later installing a static SDL2_tff)
- Allow parallel installation with cmake support of a static and shared SDL2_ttf
- Fixes CMake Macos versioning
- add MSVC to the CI test matrix
- test the installed cmake scripts

Fixes #141
Fixes #143
Fixes #226


sezero:
Does this obsolete https://github.com/libsdl-org/SDL_ttf/pull/212 ?

madebr:
> Does this obsolete #212 ?

Yes, I'll post a message there to ask for feedback.

slouken:
Go ahead and convert the PR to draft while you're iterating on it, and when you flip it back we'll merge.

slouken:
> Thanks for the trust! I'll be happy to maintain the cmake scripts and what not.

Great, thanks! :)

madebr:
Am I "allowed" to rename the cmake configure options of SDL_(image|mixer|ttf) ?
I don't know in what sense these option are "official".

In CMake, it's considered good practice/good manners to namespace your public (=cached) variables.
Right now: SDL_image has e.g. `BACKEND_WIC`, `BACKEND_STB`, `SUPPORT_AVIF` and `SUPPORT_TIF`, ...
`SDL_ttf` has: `TTF_WITH_HARFBUZZ`
`DL_mixer` has: `SUPPORT_WAV`, `SUPPORT_FLAC`, ...

I propose to prefix all options with the name of the project, that is `SDL2_IMAGE_`, `SDL2_MIXER_` and `SDL2_TTF_`.
That way, for a user, it becomes immedieately clear who is using the option.

slouken:
> Am I "allowed" to rename the cmake configure options of SDL_(image|mixer|ttf) ? I don't know in what sense these option are "official".
> 
> In CMake, it's considered good practice/good manners to namespace your public (=cached) variables. Right now: SDL_image has e.g. `BACKEND_WIC`, `BACKEND_STB`, `SUPPORT_AVIF` and `SUPPORT_TIF`, ... `SDL_ttf` has: `TTF_WITH_HARFBUZZ` `DL_mixer` has: `SUPPORT_WAV`, `SUPPORT_FLAC`, ...
> 
> I propose to prefix all options with the name of the project, that is `SDL2_IMAGE_`, `SDL2_MIXER_` and `SDL2_TTF_`. That way, for a user, it becomes immedieately clear who is using the option.

Yep, that's fine. The CMake support in those libraries has been experimental and we haven't shipped a release that anyone is depending on with CMake support.

slouken:
@madebr, I think aside from CMake support SDL_ttf is ready to build release candidates. Is there anything we're waiting on here?

madebr:
My intention was to bring SDL_image in a perfect state, and then copy everything to SDL_ttf/SDL_mixer.
(working on MSVC CI support atm)
Nothing is blocking this pr, aside from needing a rebase + application of all last-minute changes of the SDL_image cmake pr.

slouken:
> My intention was to bring SDL_image in a perfect state, and then copy everything to SDL_ttf/SDL_mixer.

That sounds great. Thanks! :)

madebr:
@slouken 
After this pr, you will have to modify your release scripts for generating the VC dev tarball, the MinGW dev tarball & the SDL2_ttf framework in the same way as you did for SDL2 and SDL2_image.

madebr:
@slouken 
Can you please test the Macos framework?
Because I don't know how the compile and link flags of SDL2, SDL2_image and SDL2_ttf will interact.

CMake has a concept of [*Option deduplication*](https://cmake.org/cmake/help/latest/command/add_compile_options.html#option-de-duplication), which we might need to avoid for the `-F` option.

FtZPetruska:
> @slouken Can you please test the Macos framework? Because I don't know how the compile and link flags of SDL2, SDL2_image and SDL2_ttf will interact.
> 
> CMake has a concept of [_Option deduplication_](https://cmake.org/cmake/help/latest/command/add_compile_options.html#option-de-duplication), which we might need to avoid for the `-F` option. Otoh, I might also worry for nothing because we modify the cmake target for xcode using `set_target_properties`.

I tested on my machine with the same setup as mentioned on [SDL2_net's PR](https://github.com/libsdl-org/SDL_net/pull/48#discussion_r895093134), and the same issue arose with option de-duplication breaking the linker flags with SDL2 framework.

slouken:
@madebr, feel free to merge this when you think it's ready.

SpaceIm:
> * Create SDL2_ttf::SDL2_ttf when building a shared library,
>   create SDL2_ttf::SDL2_ttf-static when building a static library

It's a breaking change.

PR #220, opened 2022-05-25
CHANGES.txt: Mention CVE-2022-27470

smcv:


sezero:
P.S.: Does anyone know whether the SDL-1.2 version is affected? (Not that anyone cares about it, but for completeness.)

PR #222, opened 2022-05-26
build: Compile with large inode number support where possible

smcv:
On filesystems with large inode numbers, such as overlayfs, attempting
to stat() a file on a 32-bit system using legacy syscalls can fail
with EOVERFLOW. If we opt-in to more modern "large file support"
syscalls, then source code references to functions like stat() are
transparently replaced with ABIs that support large file sizes and
inode numbers, such as stat64().

In general, SDL_ttf delegates its file I/O to SDL, so it is unaffected
by this issue. However, if we build with the vendored libraries, they
use libc functions like stat(), subjecting them to the EOVERFLOW problem.


smcv:
What this actually does behind the scenes (see `/usr/share/autoconf/autoconf/specific.m4`) can include:

* `-D_FILE_OFFSET_BITS=64` on glibc
* `-D_LARGE_FILES=1` on AIX
* `-n32` on IRIX

In the Autotools build, I think it's best to let Autoconf do what it does best, rather than doing this sort of 1990s Unix arcana directly.

In the CMake build, there's no straightforward equivalent of `AC_SYS_LARGEFILE`, but it's probably close enough if we detect glibc and use `-D_FILE_OFFSET_BITS=64`.

Issue #221, opened 2022-05-25
pysdl2 test fails after upgrading 2.0.18 to 2.19.1


smcv:
[py-sdl2](https://github.com/py-sdl/py-sdl2) has this:

```
        ret = sdlttf.TTF_SetDirection(HB_DIRECTION_LTR)
        assert ret == 0
        # Try changing the script direction to see if it has any effect
        w1, h1, w2, h2 = c_int(0), c_int(0), c_int(0), c_int(0)
        sdlttf.TTF_SizeText(font, b"Hi there!", byref(w1), byref(h1))
        ret = sdlttf.TTF_SetDirection(HB_DIRECTION_TTB)
        assert ret == 0
        sdlttf.TTF_SizeText(font, b"Hi there!", byref(w2), byref(h2))
        sdlttf.TTF_SetDirection(HB_DIRECTION_LTR) # Reset direction
        assert w1.value > w2.value
```

(This is a Python binding for SDL using ctypes, but the mapping to the C API is reasonably obvious.)

This worked with 2.0.18, but doesn't work as intended since 1b2f2d7. Previously, all font rendering used the global and script, and would change as soon as the global variables changed; but now, the font object remembers the value of the global variable at the time it was created, and subsequent changes to the global variable won't affect pre-existing font objects.

I don't know whether any actual games rely on this.

smcv:
One way to get the same semantics that the py-sdl test is checking for, while still keeping the new feature from 1b2f2d7 working, would be to treat `font->hb_direction` and `font->hb_script` as being per-font overrides for the global variables, with the default being "don't override":

* keep the global variables as-is
* initialize `font->hb_direction` and `font->hb_script` to `HB_DIRECTION_INVALID` and `HB_SCRIPT_INVALID`
* every time those struct members are used, if their value is `INVALID`, use the value of the global variable instead

sezero:
I don't _think_ the behavior change was intentional in that commit and it should be changed back, IMO. @1bsyl?

slouken:
> One way to get the same semantics that the py-sdl test is checking for, while still keeping the new feature from [1b2f2d7](https://github.com/libsdl-org/SDL_ttf/commit/1b2f2d7b142ac090519b03a8213977e091f724dc) working, would be to treat `font->hb_direction` and `font->hb_script` as being per-font overrides for the global variables, with the default being "don't override":
> 
> * keep the global variables as-is
[...]

PR #224, opened 2022-05-26
Add a semi-automatic test for #221

smcv:
This needs to be given a valid font as a command-line argument, but is
otherwise automatic.

---

I haven't wired this up to the CMake build system or the CI, because I don't want to put this on the critical path for 2.20.0, but it does seem to work.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -147,7 +147,11 @@ EXTRA_DIST = CHANGES.txt COPYING.txt README.txt \
 	SDL2_ttfConfig.cmake.in		\
 	autogen.sh gcc-fat.sh
 
-noinst_PROGRAMS = showfont glfont
+noinst_PROGRAMS = showfont glfont testfont
+
+testfont_CPPFLAGS = $(TTF_CFLAGS) $(LOCAL_HB_FLAGS)
+testfont_SOURCES = test/main.c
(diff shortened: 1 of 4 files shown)


smcv:
Given how long it took to get the SDL_image test passing reliably, if @slouken is doing 2.20.0 soon then I'm inclined to wait for 2.21.x to add this - it's just not as high a priority as the CMake build system or getting a release out.

smcv:
Given how long it took to get the SDL_image test passing reliably, if @slouken is doing 2.20.0 soon then I'm inclined to wait for 2.21.x to add this - it's just not as high a priority as the CMake build system or getting a release out.

slouken:
I'm okay with adding this now. The SDL_image test found bugs that were good to fix for the next release, so if that happens here, that makes me happy. :)

smcv:
OK, I'll try to get it integrated into CI in a way that isn't randomly regressing, but please don't block on this for 2.20.

(Also please don't imagine that this is a thorough test - it doesn't even render any text.)

slouken:
@madebr, I'm working on updating SDL_ttf 3.0, do you want to take a look at this again?

slouken:
@smcv, I'm working on a 3.0 preview release, feel free to update this PR for review and inclusion.

Issue #234, opened 2022-07-22
CMake minimum version set to 3.14 but CMakeLists.txt contains features from 3.21 


kleszcz:
While minimum required version of cmake is set to 3.14 when building shared libs ` file(COPY_FILE)` is used for installation and it's a [3.21](https://cmake.org/cmake/help/latest/command/file.html#copy-file) feature. Either bump the minimum version to 3.21 or fallback to `file(COPY)`

Tested on Ubuntu 20.04 with cmake 3.20.  Note that while using a higher version of cmake everything works and no complaint is made by cmake during generation as instructions are part of `install(CODE)` so they are not checked against cmake version.

madebr:
Fixed by https://github.com/libsdl-org/SDL_ttf/commit/c5478b14d639afc405f03c5bbdddbde58b4981bb

Please re-open if you think this solution is not sufficient.

sezero:
SDL_rtf has the same issue.

madebr:
Fixed in https://github.com/libsdl-org/SDL_rtf/commit/b5a9ad3808b19c6532ddd21edbbb07074c3625b8 (includes a bonus fix)

sezero:
Thanks

PR #235, opened 2022-07-22
Set cmake version to include file(COPY_FILE) #234

kleszcz:
As mentioned in #234 I've propose upgrading cmake minimum to 3.21

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.14)
+cmake_minimum_required(VERSION 3.21)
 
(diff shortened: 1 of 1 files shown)


madebr:
Thanks for notifying us for this issue.
CMake 3.21 is too new though because current LTS Linux distributions are shipped with a CMake version < 3.21.

In https://github.com/libsdl-org/SDL_ttf/commit/c5478b14d639afc405f03c5bbdddbde58b4981bb, I replaced the `file(COPY_FILE)` with an `execute_process(COMMAND "${CMAKE_COMMAND} -E COPY)`

PR #242, opened 2022-10-04
[Fix] Problem with DEPRECATION in CMAKE

BastienBoymond:
When you use CMake on linux this if can create bug and stop compilation of SDL2_TTF

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -395,8 +395,3 @@ add_library(SDL2::ttf INTERFACE IMPORTED GLOBAL)
 set_target_properties(SDL2::ttf PROPERTIES
     INTERFACE_LINK_LIBRARIES "SDL2_ttf"
 )
-if(CMAKE_VERSION VERSION_GREATER_EQUAL "3.17")
-    set_target_properties(SDL2::ttf PROPERTIES
-        DEPRECATION "Use SDL2_ttf::SDL2_ttf or SDL2_ttf::SDL2_ttf-static instead"
-    )
-endif()


madebr:
What is your CMake version? (output of `cmake --version`)
And what is the exact error printed?

BastienBoymond:
cmake 3.16.3

CMake Error at build/_deps/sdl2_ttf-src/CMakeLists.txt:402 (set_target_properties):
  INTERFACE_LIBRARY targets may only have whitelisted properties.  The
  property "DEPRECATION" is not allowed.

so i just deleted it

CMakeLists.txt
```diff
@@ -395,8 +395,3 @@ add_library(SDL2::ttf INTERFACE IMPORTED GLOBAL)
 set_target_properties(SDL2::ttf PROPERTIES
     INTERFACE_LINK_LIBRARIES "SDL2_ttf"
 )
-if(CMAKE_VERSION VERSION_GREATER_EQUAL "3.17")
```
madebr, line 398: This `if` conditional is supposed to protect against cmake versions < 3.17.
Please extract the [following file](https://github.com/libsdl-org/SDL_ttf/files/9711318/a.zip) to e.g. `/tmp/a.cmake`,
run it using `cmake -P /tmp/a.cmake` and post the output.

script:
```cmake
message("CMAKE_VERSION: ${CMAKE_VERSION}")
if(CMAKE_VERSION VERSION_GREATER_EQUAL "3.17.0")
  message("CMAKE_VERSION VERSION_GREATER_EQUAL 3.17: YES")
else()
  message("CMAKE_VERSION VERSION_GREATER_EQUAL: NO")
endif()
if(NOT CMAKE_VERSION VERSION_LESS "3.17.0")
  message("NOT CMAKE_VERSION VERSION_LESS 3.17: YES")
else()
  message("NOT CMAKE_VERSION VERSION_LESS 3.17: NO")
endif()
```

For my system, the output is:
```sh
$ cmake -P /tmp/a.cmake
CMAKE_VERSION: 3.22.2
CMAKE_VERSION VERSION_GREATER_EQUAL 3.17: YES
NOT CMAKE_VERSION VERSION_LESS 3.17: YES
```

BastienBoymond: for my system was that
`CMAKE_VERSION: 3.16.3
CMAKE_VERSION VERSION_GREATER_EQUAL: NO
NOT CMAKE_VERSION VERSION_LESS 3.17: NO`

But the problem was in the 
`set_target_properties` that use DEPRECATED and was not allowed.

madebr: > for my system was that `CMAKE_VERSION: 3.16.3 CMAKE_VERSION VERSION_GREATER_EQUAL: NO NOT CMAKE_VERSION VERSION_LESS 3.17: NO`

This output is correct for CMake 3.16.3.

> But the problem was in the `set_target_properties` that use DEPRECATED and was not allowed.

The [`DEPRECATION` property](https://cmake.org/cmake/help/v3.17/prop_tgt/DEPRECATION.html) has been available since CMake 3.17.
So the `set_target_properties` line should not have been reached.

I tried [CMake 3.16.3 for Linux](https://cmake.org/files/v3.16/cmake-3.16.3-Linux-x86_64.tar.gz) and the last lines of configuring this project with `cmake --trace` are:
```
/home/maarten/programming/SDL_ttf/CMakeLists.txt(398):  if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.17 )
-- Configuring done
-- Generating done
-- Build files have been written to: /tmp/build
```
==> The `set_target_properties` inside the `if` branch is skipped.

madebr:
What is your os and what is the origin of your cmake installation?
Perhaps it can be reproduced in a docker container.

BastienBoymond:
Ubuntu 20
the origin was a project who will use sdl sdl_image and sdl_ttf

madebr:
Closing this pr.

If somebody is able to reproduce this, please reply with logs and how to do this.

PR #243, opened 2022-10-20
cmake: make cmake installation folder configurable

madebr:
Follow-up of https://github.com/libsdl-org/SDL/pull/6415


CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.14)
+cmake_minimum_required(VERSION 3.16)
```
madebr, line 1: This is the same minimum required version as all other SDL_xxx libraries.

PR #251, opened 2023-01-10
Use SDL_aligned_alloc() instead of relying on SDL internal implementa…

slouken:
…tion


1bsyl:
yes seems ok ! merging and I will double-chech!

1bsyl:
double check is ok.
I just still wonder if we need 
https://github.com/libsdl-org/SDL_ttf/blob/main/SDL_ttf.c#L1399
`        SDL_size_add_overflow(pitch, alignment, &pitch) ||`
I removed it before, but that was maybe a mistake.

1bsyl:
so removing it https://github.com/libsdl-org/SDL_ttf/commit/d9037f526a18a11766804149e52be7eca0ce83bd
it doesn't crash after millions of random generation.
(removing 1 more byte would crash after ~50 generation)

slouken:
> so removing it [d9037f5](https://github.com/libsdl-org/SDL_ttf/commit/d9037f526a18a11766804149e52be7eca0ce83bd) it doesn't crash after millions of random generation. (removing 1 more byte would crash after ~50 generation)

Oh yeah, that wasn't needed, thanks!

slouken:
Also, I wonder if we need this:
```
    pitch &= ~alignment;
```

1bsyl:
yes, tested, because destination need to be aligned

slouken:
I think that's probably why the addition of alignment was needed:
```c
    if (width > SDL_MAX_SINT32 ||
        height > SDL_MAX_SINT32 ||
        SDL_size_add_overflow(width, alignment, &pitch) ||
        SDL_size_mul_overflow(pitch, bytes_per_pixel, &pitch) ||
        SDL_size_add_overflow(pitch, alignment, &pitch) ||
        pitch > SDL_MAX_SINT32) {
        return NULL;
    }
    pitch &= ~alignment;
```

The final statement can shrink pitch by up to alignment bytes, but we shouldn't shrink it below width * alignment * bytes_per_pixel. I think the fact that alignment is a factor in the previous multiplication means that `pitch & alignment` always equals 0, so that statement has no effect. If that's not true, then I think we still need the addition of alignment to pitch. Can you verify?

1bsyl:
yes, I ll look back at the code deeper later :)

slouken:
Oh, I take it back, alignment isn't a factor, the calculation is actually (width + alignment) * bytes_per_pixel

So in the case where width = 2, height = 2, alignment = 15, bytes_per_pixel = 1, pitch would equal 17, then truncated to 15, which is I guess what we want here. Do we actually write up to alignment pixels beyond the edge of the surface?

1bsyl:
I've double check and this ok ! 
some explanation: https://github.com/libsdl-org/SDL_ttf/commit/02f9def1b692837672425b77be928707ccd75a6d

in short,in the fast mode (SSE, NEON), it write blocks by blocks, at block-aligned addresses. with a ratio 1:byte_per_pixel.
(for the blended mode, read 1 pixel, then  write a pixel of 4 bytes). blocks have the size of sse 128. eg 16 bytes. (alignement :)
so this solve the worst case when we write at last pixel width.

( I also notice, maybe I did before, that SDL_malloc aligned is extra padded. so it makes it harder to detect out of bounds with valgrind. _

Issue #98, opened 2021-02-11
SDL_ttf test app + patches to apply to FreeType-2.9.1


SDLBugzilla:
# This bug report was migrated from our old Bugzilla tracker.

These attachments are available in the static archive:

* ~~[test app for SDL_ttf (SDL_ttf_testapp.c, text/x-csrc, 2018-11-09 15:38:50 +0000, 25211 bytes)](https://bugzilla.libsdl.org/attachment.cgi?id=3480)~~
* [fonts (font.tar.gz, application/gzip, 2018-11-09 15:41:33 +0000, 9812156 bytes)](https://bugzilla.libsdl.org/attachment.cgi?id=3481)
* ~~[test app for SDL_ttf (SDL_ttf_testapp.c, text/x-csrc, 2018-11-13 10:07:23 +0000, 27734 bytes)](https://bugzilla.libsdl.org/attachment.cgi?id=3485)~~
* ~~[test app for SDL_ttf (SDL_ttf_testapp.c, text/x-csrc, 2018-11-14 12:01:32 +0000, 27827 bytes)](https://bugzilla.libsdl.org/attachment.cgi?id=3493)~~
* ~~[test app for SDL_ttf (SDL_ttf_testapp.c, text/x-csrc, 2018-11-26 15:09:21 +0000, 32231 bytes)](https://bugzilla.libsdl.org/attachment.cgi?id=3505)~~
* [test app for SDL_ttf (SDL_ttf_testapp.c, text/x-csrc, 2018-12-13 16:04:29 +0000, 37691 bytes)](https://bugzilla.libsdl.org/attachment.cgi?id=3546)

**Reported in version:** unspecified
**Reported for operating system, platform:** Linux, x86_64

# Comments on the original bug report:

On 2018-11-09 15:38:50 +0000, Sylvain wrote:

> Created attachment 3480
> test app for SDL_ttf
> 
> Here's the console test app I've used for last bugs and also as a text rendering stress-test.
> 
> You can specify several fonts, strings, and interacts with the keyboard:
> - press several times "space" and you go with all strings and render modes (solid,shaded,blended).
> - use keyboard to: switch font, size, style, outline, kerning, hinting, wrap, etc. F1 for help!
> 
> I've added a few known broken strings and random strings. And also all fonts I found in SDL_ttf bug tracker (scalable/non-scalable. Extension .ttf, .fon, .pcf).
> 
[...]

PR #256, opened 2023-01-16
Add test app

1bsyl:
Add test app / fuzzer for SDL_ttf, 
this the test app from https://github.com/libsdl-org/SDL_ttf/issues/98
that I use for find crash / bug with SDL_ttf

### DIFF
--- a/testapp.c
+++ b/testapp.c
@@ -1,0 +1,1240 @@
+/*
+  testapp:  An example of using the SDL_ttf library with OpenGL.
+  Copyright (C) 2001-2023 Sam Lantinga <slouken@libsdl.org>
+
+  This software is provided 'as-is', without any express or implied
+  warranty.  In no event will the authors be held liable for any damages
+  arising from the use of this software.
+
+  Permission is granted to anyone to use this software for any purpose,
+  including commercial applications, and to alter it and redistribute it
+  freely, subject to the following restrictions:
+
+  1. The origin of this software must not be misrepresented; you must not
+     claim that you wrote the original software. If you use this software
+     in a product, an acknowledgment in the product documentation would be
+     appreciated but is not required.
+  2. Altered source versions must be plainly marked as such, and must not be
+     misrepresented as being the original software.
+  3. This notice may not be removed or altered from any source distribution.
+*/
+
+/* A simple program to test the text rendering feature of the TTF library,
+ * and find bugs
+ *
+ * <F1> for console help
+ *
+ * Use
+ * - fill/replace test_fonts[] with your fonts
+ * - replay a case, adding the log at line ~650 (search for 'replay')
+ * - add more string to test_strings[]
+ *
+ */
+
+const static int g_force_no_SDF = 0; /* make random fuzzer faster by disabling SDF rendering */
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <time.h>
+#include <libgen.h>
+
+#include "SDL3/SDL.h"
+#include "SDL3/SDL_ttf.h"
+
+
+static void help(void)
+{
+    SDL_Log("up/down : font size -/+");
+    SDL_Log("c/v : next/previous font");
+    SDL_Log("o/p : outline   -/+");
+    SDL_Log("q/e : wrap size -/+");
+    SDL_Log("a : align wrap: left center right");
+    SDL_Log("w   : wrap");
+    SDL_Log("u   : underline");
+    SDL_Log("k   : kerning");
+    SDL_Log("s   : strike-through");
+    SDL_Log("b   : bold");
+    SDL_Log("g/h : hinting -/+ (normal, light, light_subpix, mono, none)");
+    SDL_Log("i   : italic");
+    SDL_Log("f   : signed distance field");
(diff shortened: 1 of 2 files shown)


testapp.c
```diff
@@ -0,0 +1,1232 @@
...
+/* RENDER_SOLID = 0, RENDER_BLENDED = 1, RENDER_SHADED = 2, RENDER_LCD = 3 } */
+static int render_mode = -1;
+static int render_mode_overwrite;
+static int render_mode_count = 4;
+static const char *render_mode_desc[] = { "Solid", "Blended", "Shaded", "LCD" };
+
+static int direction = 0;
+static int direction_count = 4;  
+static const char *direction_desc[] = { "LTR", "RTL", "TTB", "BTT"  };
+static const int  *direction_val[] = { TTF_DIRECTION_LTR, TTF_DIRECTION_RTL, TTF_DIRECTION_TTB, TTF_DIRECTION_BTT };
```
madebr, line 164:
```suggestion
static const int   direction_val[] = { TTF_DIRECTION_LTR, TTF_DIRECTION_RTL, TTF_DIRECTION_TTB, TTF_DIRECTION_BTT };
```

1bsyl: good catch ! done

madebr:
Could you upload a new fonts archive somewhere?
Because the one from https://github.com/libsdl-org/SDL_ttf/issues/98#issue-806097965 does not contain `NotoColorEmoji.ttf` (and possibly more).

1bsyl:
not sure if we need to provide fonts.
this can be used with any font. 
there was an idea to scan a "./fonts" to find fonts.
but sdl doesn't provide directory listing (yet)

testapp.c
```diff
@@ -0,0 +1,1232 @@
...
+#if 0
+# include "fonts/fuzzing_corpora.h"
+#endif
+#if 0
+# include "fonts/android_fonts_others.h"
+# include "fonts/android_fonts.h"
+#endif
+    "fonts/Kaumudi.ttf"
+};
+static const int test_fonts_count = sizeof(test_fonts)/sizeof(char*);
```
madebr, line 132: You can use `SDL_arraysize` here, and in other locations.

testapp.c
```diff
@@ -0,0 +1,1232 @@
...
+/* RENDER_SOLID = 0, RENDER_BLENDED = 1, RENDER_SHADED = 2, RENDER_LCD = 3 } */
+static int render_mode = -1;
+static int render_mode_overwrite;
+static int render_mode_count = 4;
+static const char *render_mode_desc[] = { "Solid", "Blended", "Shaded", "LCD" };
+
+static int direction = 0;
+static int direction_count = 4;  
+static const char *direction_desc[] = { "LTR", "RTL", "TTB", "BTT"  };
+static const int  direction_val[] = { TTF_DIRECTION_LTR, TTF_DIRECTION_RTL, TTF_DIRECTION_TTB, TTF_DIRECTION_BTT };
```
madebr, line 164: I would do this as:
```c
static struct {
    const char *description;
    int value;
} directions[] = {
    { "LTR", TTF_DIRECTION_LTR },
    { "RTL", TTF_DIRECTION_RTL },
    { "TTB", TTF_DIRECTION_TTB },
    { "BTT", TTF_DIRECTION_BTT }
};
```
And use `SDL_arraysize` for `direction_count`.

1bsyl: yes done!
this was a test app that evolved occasionally,
there is definitively some room for improvement !

testapp.c
```diff
@@ -0,0 +1,1240 @@
...
+    { "RTL", TTF_DIRECTION_RTL },
+    { "TTB", TTF_DIRECTION_TTB },
+    { "BTT", TTF_DIRECTION_BTT }
+};
+static int direction_count = SDL_arraysize(directions);
+
+
+//static const char *hinting_desc[] = { "normal", "light", "light_subpix", "lcd_subpix", "mono", "none" };
+static const char *hinting_desc[] = { "normal", "light", "light_subpix", "mono", "none" };
+static int hinting_count = SDL_arraysize(hinting_desc);
```
madebr, line 176:
```suggestion
static const int hinting_count = SDL_arraysize(hinting_desc);
```

testapp.c
```diff
@@ -0,0 +1,1240 @@
...
+static int seed = 0;
+static int print_elapsed_ticks = 0;
+static int update_screen_mode = 0;
+static int save_to_bmp = 0;
+
+/* RENDER_SOLID = 0, RENDER_BLENDED = 1, RENDER_SHADED = 2, RENDER_LCD = 3 } */
+static int render_mode = -1;
+static int render_mode_overwrite;
+static const char *render_mode_desc[] = { "Solid", "Blended", "Shaded", "LCD" };
+static int render_mode_count = SDL_arraysize(render_mode_desc);
```
madebr, line 159:
```suggestion
static const int render_mode_count = SDL_arraysize(render_mode_desc);
```

1bsyl: updated!

1bsyl:
Re-based and fixed build of testapp.c

PR #268, opened 2023-03-04
Add TTF_GetFontScalable to distinguish bitmap vs outline fonts

### LINKED ISSUE #169: Add bitmap font detection function
Freetype supports COLR/CPAL/SVG (vector based) as well as CBDT/CBLC (bitmap based png's) fonts. However I cannot find a function in SDL_ttf API that can tell me if the font is bitmap or not. If it doesn't exist yet - it should. Am I missing something?

MyreMylar:
closes #169

This is the freetype function used for bitmap font detection in pygame's `freetype` submodule. I've been recently looking at how feasible it is to replace the old `pygame.freetype` submodule with just our SDL_ttf wrapper `pygame.font` and one of the snags I noticed is bitmap font detection.

I looked over here and saw the issue so thought I'd have a stab at a PR.

My first SDL_ttf PR so I've almost certainly missed a bunch of things everyone is supposed to do. So, sorry about that.

### DIFF
--- a/include/SDL3/SDL_ttf.h
+++ b/include/SDL3/SDL_ttf.h
@@ -2311,6 +2311,21 @@ extern DECLSPEC int SDLCALL TTF_SetFontDirection(TTF_Font *font, TTF_Direction d
  */
 extern DECLSPEC int SDLCALL TTF_SetFontScriptName(TTF_Font *font, const char *script);
 
+/**
+ * Query whether a font is scalable or not.
+ *
+ * Scalability lets us distinguish between outline and bitmap fonts.
+ *
+ * \param font the font to query
+ *
+ * \returns SDL_TRUE if the font is scalable, SDL_FALSE otherwise.
+ *
+ * \since This function is available since SDL_ttf 2.21.0.
+ *
+ * \sa TTF_SetFontSDF
+ */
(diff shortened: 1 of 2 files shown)


1bsyl:
seems ok to me. this was asked before! @slouken

slouken:
This seems fine to me. I would probably rename the function `TTF_IsFontScalable()` to indicate it's returning a boolean value.

slouken:
Merged, thanks!

sezero:
Do we want this in the SDL2 branch too?

slouken:
No, this will be an SDL3 API

include/SDL3/SDL_ttf.h
```diff
@@ -2311,6 +2311,21 @@ extern DECLSPEC int SDLCALL TTF_SetFontDirection(TTF_Font *font, TTF_Direction d
...
+/**
+ * Query whether a font is scalable or not.
+ *
+ * Scalability lets us distinguish between outline and bitmap fonts.
+ *
+ * \param font the font to query
+ *
+ * \returns SDL_TRUE if the font is scalable, SDL_FALSE otherwise.
+ *
+ * \since This function is available since SDL_ttf 2.21.0.
```
Starbuck5, line 2323: Since this is only going into SDL_ttf 3, I believe this \since label needs to be revised.

PR #277, opened 2023-03-26
Don't add -Wl,--no-undefined for clang, windows 10

AlpRabbit:
Since clang on Windows 10 doesn't have `-Wl,--no-undefined` option, disabling it. [Same thing is done in SDL](https://github.com/libsdl-org/SDL/commit/34ecd71e80b651c9840f60d91be60e30305fdc6f).


sezero:
The real problem is why is the cmake test not catching that that linker option isn't supported, not just here but also in SDL itself?

madebr:
Looks like SDL_ttf is not calling this function.
This piece of code is missing in the main `CMakeLists.txt`:
```cmake
if(SDL3IMAGE_BUILD_SHARED_LIBS)
    sdl_target_link_options_no_undefined(SDL3_ttf)
endif()
```

Like [here](https://github.com/libsdl-org/SDL_image/blob/1ccf79dd76af3a4d9f5d5259a3083a84ac519fd3/CMakeLists.txt#L288-L290).

madebr (approved):

Issue #279, opened 2023-03-28
Generated file has incorrect include directories


markand:
I think the CMake generated file as an invalid include directories because once you try to link to `SDL3_ttf::SDL3_ttf` it ends up with:

```
-- Configuring done (0.1s)
CMake Error in src/CMakeLists.txt:
  Imported target "SDL3_ttf::SDL3_ttf" includes non-existent path

    "/include"

  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:

  * The path was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.

  * The installation package was faulty and references files it does not
  provide.



-- Generating done (0.0s)
CMake Generate step failed.  Build files cannot be regenerated correctly.
```

This can be experienced using a simple:

```cmake
find_package(SDL3_ttf REQUIRED)
target_link_libraries(myapp SDL3_ttf::SDL3_ttf)
```

In the shared-targets.cmake file there is:

```
set_target_properties(SDL3_ttf::SDL3_ttf PROPERTIES
  COMPATIBLE_INTERFACE_BOOL "SDL3_SHARED"
  INTERFACE_INCLUDE_DIRECTORIES "/include;/include/SDL3;${_IMPORT_PREFIX}/include;${_IMPORT_PREFIX}/include/SDL3"      
                                                                    ^^^^^^^^^^^^^^^^^^^   
  INTERFACE_SDL3_SHARED "ON"
)
```

I'm digging why cause that issue.

madebr:
Fixed by #280

PR #280, opened 2023-03-28
Fix CMake generator expression

markand:
This fixes issue #279 because the `/include` part is outside of the generator expression.


madebr (approved):

madebr:
Odd, only SDL_ttf had this bug.
Thanks for fixing this!

Issue #290, opened 2023-07-11
In Visual Studio projects SDL_ttf.h is not directly accessible


PurityLake:
This is a rather small issue but when using the Visual Studio projects, when you attempt to open `SDL_ttf.h` it complains that the file does not exists.

This is just a configuration problem so it is an easy fix.

madebr:
Fixed by #291. Thanks again!

sezero:
Fixed by https://github.com/libsdl-org/SDL_ttf/pull/291 ?

sezero:
> Fixed by #291 ?

Was too late :)

madebr:
I :gun:  first :P

PR #291, opened 2023-07-11
Fixed Visual Studio project files with incorrect paths

PurityLake:
#290

This simply fixes the `.vcxproject` and `.vcxproject.filters` to point to the correct file.

It was previously just pointing to `..\SDL_ttf.h`, I assume this was the case at one point in time but it has seeming changed. It now points to `..\include\SDL3\SDL_ttf.h`.


madebr (approved):

madebr:
Thanks for the fixes!

sezero:
Off-topic: can we change SDL3_ttf CI to use setup-sdl like others?

madebr:
I just merged https://github.com/libsdl-org/SDL_ttf/pull/289

PR #293, opened 2023-07-14
Use SDL3_ttf_0.0.0 for versioned symbols, not SDL3_mixer_0.0.0

smcv:
I assume this was a copy/paste mistake.

### DIFF
--- a/SDL_ttf.sym
+++ b/SDL_ttf.sym
@@ -1,4 +1,4 @@
-SDL3_mixer_0.0.0 {
+SDL3_ttf_0.0.0 {
   global:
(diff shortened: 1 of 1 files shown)


madebr:
This was my bad. Thanks for noticing and fixing.

smcv:
FYI I'm starting to look at packaging the SDL3 satellite libraries in Debian experimental, which is why I'm noticing this sort of thing. (I know they're not stable APIs yet, but in Debian experimental there's no expectation of things working reliably - it's for developers only, and breaking changes are considered to be OK.)

Issue #308, opened 2023-10-07
[Feature request] More text-shaping functionality.


LucidSigma:
## Font language.
This library currently has support for setting a font's script (via `TTF_SetFontScriptName`) and text direction (via `TTF_SetFontDirection`) to assist with text-shaping using HarfBuzz. However, HarfBuzz has another shaping setting — which is language. This is done with [hb_buffer_set_language](https://harfbuzz.github.io/harfbuzz-hb-buffer.html#hb-buffer-set-language).

A text's language is separate from its script. For example: Arabic, Persian, Urdu, and Sindhi all use an Arabic script, but have different alphabets and shaping rules.

I suggest the following API (to be congruent with the aforementioned functions):
```c
/*
    font: the font to specify the language for.
    language_bcp47: a null-terminated string containing the desired language's BCP47 code.
    returns: 0 on success, -1 on error.
*/
int TTF_SetFontLanguage(TTF_Font *font, const char *language_bcp47);
```

## Font beginning/end flags.
Some more advanced shaping features can be set using [hb_buffer_set_flags](https://harfbuzz.github.io/harfbuzz-hb-buffer.html#hb-buffer-set-flags). The list of available flags can be found [here](https://harfbuzz.github.io/harfbuzz-hb-buffer.html#hb-buffer-flags-t). As mentioned, most of these are advanced and out-of-scope for SDL_ttf; however I do feel that `HB_BUFFER_FLAG_BOT` and `HB_BUFFER_FLAG_EOT` could be implemented in SDL_ttf. These flags basically say that the shaping buffer's text is at the beginning/end of a paragraph (as some languages — such as Arabic — have specific text-shaping rules in these contexts).

Similarly to the previous suggestion, here is what the example API could look like herefor:
```c
/*
    font: the font to specify the beginning flag for.
    is_beginning: 0 to specify not at beginning of paragraph, non-zero otherwise.
*/
void TTF_SetFontBeginningOfParagraph(TTF_Font *font, int is_beginning);

/*
    font: the font to specify the end flag for.
    is_end: 0 to specify not at end of paragraph, non-zero otherwise.
*/
void TTF_SetFontEndOfParagraph(TTF_Font *font, int is_end);
```

1bsyl:
If this does better shaping, then yes, we should provide this.
I can create a PR

1bsyl:
Using the correct language (in addition to script), change a little bit the glyph/position for some language (and it's said to be orthogonal to script).

[...]

PR #310, opened 2023-10-09
Add TTF_SetFontLanguage to set language BCP47 code, (see #308)

1bsyl:
Add TTF_SetFontLanguage to set language BCP47 code, (see #308)

### DIFF
--- a/src/SDL_ttf.c
+++ b/src/SDL_ttf.c
@@ -292,6 +292,7 @@ struct _TTF_Font {
     hb_script_t hb_script;
     /* If HB_DIRECTION_INVALID, use global default direction */
     hb_direction_t hb_direction;
+    hb_language_t hb_language;
 #endif
     int render_sdf;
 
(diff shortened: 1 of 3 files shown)


src/SDL_ttf.c
```diff
@@ -3125,6 +3127,24 @@ int TTF_SetFontScriptName(TTF_Font *font, const char *script)
...
+    if (language_bcp47 == NULL) {
+        font->hb_language = hb_language_from_string("", -1);
+    } else {
+        font->hb_language = hb_language_from_string(language_bcp47, -1);
+    }
+    return 0;
+#else
+    (void) font;
+    (void) language_bcp47;
+    return -1;
```
madebr, line 3144: It looks like SDL_ttf does not set errors for unimplemented functionality. Shouldn't it?

```suggestion
    TTF_SetError("unsupported");
    return -1;
```

1bsyl: done (1 line + upper case) ! thanks!

PR #312, opened 2023-10-26
Add TTF_GetScript to retreive the script to which unicode character belongs

j9hnny97:
At the moment **SDL_ttf** provides _TTF_SetFontScriptName_, a **Harfbuzz** related function that allows a script to be set (for a specific font) to be used for text shaping.

This is useful as long as the client knows the origin of its texts, allowing him to simply set the script relative to his language. If this isn't the case how would he know what to script to use? And what would happen if there are different characters in his text, each one requiring different scripts?

**Harfbuzz** already provides a way to obtain the best script to be used by each character. This PR exposes this functionally with _TTF_GetScript_, allowing the client to deal with this issue however he desires, without compromising the current library implementation

### DIFF
--- a/SDL_ttf.c
+++ b/SDL_ttf.c
@@ -3128,6 +3128,46 @@ int TTF_SetFontScriptName(TTF_Font *font, const char *script)
 #endif
 }
 
+int TTF_GetScriptName(Uint32 ch, char *script, size_t script_size)
+{
+#if TTF_USE_HARFBUZZ
+
+    TTF_CHECK_POINTER(script, -1);
+    if (script_size < 5) {
+        return TTF_SetError("Insufficient script buffer size");
+    }
+
+    hb_buffer_t* hb_buffer = hb_buffer_create();
+
+    if (hb_buffer == NULL) {
+        return TTF_SetError("Cannot create harfbuzz buffer");
+    }
+
+    hb_unicode_funcs_t* hb_unicode_functions = hb_buffer_get_unicode_funcs(hb_buffer);
+
+    if (hb_unicode_functions == NULL) {
+        hb_buffer_destroy(hb_buffer);
+        return TTF_SetError("Cannot get harfbuzz unicode funcs");
+    }
+
+    hb_buffer_clear_contents(hb_buffer);
+    hb_buffer_set_content_type(hb_buffer, HB_BUFFER_CONTENT_TYPE_UNICODE);
+
+    uint8_t const untagged_script[4] = { HB_UNTAG(hb_unicode_script(hb_unicode_functions, ch)) };
+    script[0] = (char)untagged_script[0];
+    script[1] = (char)untagged_script[1];
+    script[2] = (char)untagged_script[2];
+    script[3] = (char)untagged_script[3];
+    script[4] = '\0';
+
+    hb_buffer_destroy(hb_buffer);
+    return 0;
+
+#else
+    return TTF_SetError("Unsupported");
+#endif
+}
+
 static int TTF_Size_Internal(TTF_Font *font,
         const char *text, const str_type_t str_type,
         int *w, int *h, int *xstart, int *ystart,
--- a/SDL_ttf.h
+++ b/SDL_ttf.h
@@ -2311,6 +2311,24 @@ extern DECLSPEC int SDLCALL TTF_SetFontDirection(TTF_Font *font, TTF_Direction d
  */
 extern DECLSPEC int SDLCALL TTF_SetFontScriptName(TTF_Font *font, const char *script);
 
+/**
+ * Query the script to which unicode character belongs.
+ *
+ * The supplied script pointer should be able to hold four characters and
+ * the null-terminator.
+ *
+ * If SDL_ttf was not built with HarfBuzz support, this function returns -1.
+ *
+ * \param ch the character to check.
+ * \param script on return, filled in with the the script as a null-terminated
+ *               string of exactly 4 characters
+ * \param script_size size of the script buffer, must be at least 5 (see above.)
+ * \returns 0 on success, or -1 on error.
+ *
+ * \since This function is available since SDL_ttf 2.23.0.
+ */
+extern DECLSPEC int SDLCALL TTF_GetScriptName(Uint32 ch, char *script, size_t script_size);
+
 /* Ends C function definitions when using C++ */
 #ifdef __cplusplus
 }


SDL_ttf.c
```diff
@@ -3128,6 +3128,44 @@ int TTF_SetFontScriptName(TTF_Font *font, const char *script)
...
+    uint8_t const untagged_script[4] = { HB_UNTAG(hb_unicode_script(hb_unicode_functions, ch)) };
+
+    script[0] = (char)untagged_script[0];
+    script[1] = (char)untagged_script[1];
+    script[2] = (char)untagged_script[2];
+    script[3] = (char)untagged_script[3];
+
+    hb_buffer_destroy(hb_buffer);
+    return 0;
+
```
slouken, line 3163: This should be:
```c
#else
    TTF_SetError("Unsupported");
    return -1;
#endif
```

SDL_ttf.c
```diff
@@ -3128,6 +3128,44 @@ int TTF_SetFontScriptName(TTF_Font *font, const char *script)
...
+extern DECLSPEC int SDLCALL TTF_GetScript(Uint32 ch, char* script)
+{
+#if TTF_USE_HARFBUZZ
+
+    hb_buffer_t* hb_buffer = hb_buffer_create();
+
+    if (hb_buffer == NULL)
+    {
+        TTF_SetError("Cannot create harfbuzz buffer");
+        return -1;
```
slouken, line 3140: The error paths should destroy hb_buffer.

slouken: Well, not this one, but the next one should.

SDL_ttf.c
```diff
@@ -3128,6 +3128,44 @@ int TTF_SetFontScriptName(TTF_Font *font, const char *script)
 #endif
 }
 
+extern DECLSPEC int SDLCALL TTF_GetScript(Uint32 ch, char* script)
```
slouken, line 3131: This function should probably take a `scriptlen` parameter and document that it should be at least 5 characters long and then null terminate the string.

j9hnny97: I don't think so. The script is always 4 characters long. I've added a check to cover that just like it happens in TTF_SetFontScriptName.

1bsyl: actually that always 4 char, so maybe "char script[4]" (or 5 to have it null terminated)

j9hnny97: @1bsyl TTF_SetFontScriptName does exactly the same thing... why not keep it consistent?

1bsyl: yes this is right ... that was just a suggestion.

SDL_ttf.c
```diff
@@ -3128,6 +3128,52 @@ int TTF_SetFontScriptName(TTF_Font *font, const char *script)
...
+
+    hb_buffer_t* hb_buffer = hb_buffer_create();
+
+    if (hb_buffer == NULL)
+    {
+        TTF_SetError("Cannot create harfbuzz buffer");
+        return -1;
+    }
+
+    if (script == NULL || SDL_strlen(script) != 4)
```
sezero, line 3143: `strlen` here doesn't make any sense at all

SDL_ttf.c
```diff
@@ -3128,6 +3128,52 @@ int TTF_SetFontScriptName(TTF_Font *font, const char *script)
...
+
+    hb_buffer_clear_contents(hb_buffer);
+    hb_buffer_set_content_type(hb_buffer, HB_BUFFER_CONTENT_TYPE_UNICODE);
+
+    uint8_t const untagged_script[4] = { HB_UNTAG(hb_unicode_script(hb_unicode_functions, ch)) };
+
+    script[0] = (char)untagged_script[0];
+    script[1] = (char)untagged_script[1];
+    script[2] = (char)untagged_script[2];
+    script[3] = (char)untagged_script[3];
```
sezero, line 3166: Do we not want nul terminating `script` here?

SDL_ttf.h
```diff
@@ -2311,6 +2311,21 @@ extern DECLSPEC int SDLCALL TTF_SetFontDirection(TTF_Font *font, TTF_Direction d
  */
 extern DECLSPEC int SDLCALL TTF_SetFontScriptName(TTF_Font *font, const char *script);
 
+/**
+ * Retrieves the script to which unicode character belongs.
+ *
+ * The supplied script value must be exactly four characters.
```
sezero, line 2317: bad documentation: this new procedure _retrieves_ the script value. Should be something like `The supplied script pointer should be able to hold four characters.` And if we want to null-terminate it, adjust it accordingly, e.g. like `The supplied script pointer should be able to hold four characters and the null-terminator.`

SDL_ttf.c
Magueija: Resolved

SDL_ttf.c
Magueija: Resolved, used TTF_CHECK_POINTER() macro instead

SDL_ttf.h
Magueija: Resolved, added on the description

sezero:
Better after the latest changes. I suggest the following to accomodate @slouken's arguments at https://github.com/libsdl-org/SDL_ttf/pull/312#discussion_r1549876730: it makes it safer, just in case. Also changes some error returns to simply return TTF_SetError(), because TTF_SetError (which is SDL_SetError) already returns -1 (there are other places in existing code that needs a similar change, but that's out of the scope of this PR.)

```diff
diff --git a/SDL_ttf.c b/SDL_ttf.c
index d6f6301..54d3575 100644
--- a/SDL_ttf.c
+++ b/SDL_ttf.c
@@ -3143,25 +3143,25 @@ int TTF_SetFontScriptName(TTF_Font *font, const char *script)
 #endif
 }
 
-extern DECLSPEC int SDLCALL TTF_GetScript(Uint32 ch, char script[5])
+int TTF_GetScript(Uint32 ch, char *script, size_t bufsize)
 {
 #if TTF_USE_HARFBUZZ
 
     TTF_CHECK_POINTER(script, -1);
+    if (bufsize < 5) {
+        return TTF_SetError("Insufficient script buffer size");
+    }
 
     hb_buffer_t* hb_buffer = hb_buffer_create();
-
     if (hb_buffer == NULL) {
-        TTF_SetError("Cannot create harfbuzz buffer");
-        return -1;
+        return TTF_SetError("Cannot create harfbuzz buffer");
     }
 
     hb_unicode_funcs_t* hb_unicode_functions = hb_buffer_get_unicode_funcs(hb_buffer);
 
     if (hb_unicode_functions == NULL) {
-        TTF_SetError("Cannot get harfbuzz unicode funcs");
         hb_buffer_destroy(hb_buffer);
-        return -1;
+        return TTF_SetError("Cannot get harfbuzz unicode funcs");
     }
 
     hb_buffer_clear_contents(hb_buffer);
@@ -3179,8 +3179,7 @@ extern DECLSPEC int SDLCALL TTF_GetScript(Uint32 ch, char script[5])
     return 0;
 
 #else
-    TTF_SetError("Unsupported");
-    return -1;
+    return TTF_SetError("Unsupported");
 #endif
 }
 
diff --git a/SDL_ttf.h b/SDL_ttf.h
index 03b9045..825bc08 100644
--- a/SDL_ttf.h
+++ b/SDL_ttf.h
@@ -2333,11 +2333,14 @@ extern DECLSPEC int SDLCALL TTF_SetFontScriptName(TTF_Font *font, const char *sc
  * \param ch the character to check.
  * \param script on return, filled in with the the script as a null-terminated
  *               string of exactly 4 characters
+ * \param bufsize size of the script buffer size, must be at least 5 in order
+                  to hold the script as a null-terminated string of exactly 4
+                  characters.
  * \returns 0 on success, or -1 on error.
  *
- * \since This function is available since SDL_ttf ?.
+ * \since This function is available since SDL_ttf 2.24.0.
  */
-extern DECLSPEC int SDLCALL TTF_GetScript(Uint32 ch, char script[5]);
+extern DECLSPEC int SDLCALL TTF_GetScript(Uint32 ch, char *script, size_t bufsize);
 
 /* Ends C function definitions when using C++ */
 #ifdef __cplusplus
```

After you apply this, maintainers will review and decide, I guess.

sezero:
This looks fair to me now. @1bsyl, @slouken: what do you think?

1bsyl:
look ok to me! 
minor suggestion, since we have the function "TTF_SetFontScriptName()", maybe this should be called TTF_GetScript**Name**()


also what is the use case ? check the script of a string by the first char ? 
If we want to validate a full string, we need a big loop, with per char, an hb_buffer_create/destroy().
Also usefull, could be an helper function to parse a long string, return an array a scripts and indexes of sub strings, for each string that has different scripts.
Or a function that get the script of a string and advance into the string as much as possible (behavior like strstr()).

Magueija:
Hi @1bsyl,

The function has the name "TTF_GetScript()" since we also have the "TTF_SetScript()" and because it is not related with fonts as "TTF_SetFontScriptName()". However, if you still think we should add "Name" at the end, I am ok with it.

Regarding the use of it:
Each character has a defined script. 'A' belongs to the Latin script, while 'ب' belongs to the Arab script, but there are some characters that could be rendered with more than one script, it just depends on the context, for example, the application current language, and, on these cases, it should be decided by the application.
Actually there can be the case where the application could render the whole string just depending on the first character, as is recommended on the unicode.org. But this does not cover all cases, some other applications could want to render the string with multiple scripts (applying them for each chunk) because they have multi-language strings.

If we want to validate a full string, we need a big loop indeed, but if we leave it to the application, now it has the possibility to do some other things on this loop and avoids a loop on the SDL_TTF and another loop on the application.

Again, having a function like strstr() seems a good ideia, but how do we decide what to do on the special cases? Which scripts should we choose?

See more: https://www.unicode.org/reports/tr24/

1bsyl:
@Magueija Are you sure, I can't find TTF_SetScript ?!
otherwise I am ok with anything

Magueija:
@1bsyl I just noticed it's deprecated, I will add suffix.
https://wiki.libsdl.org/SDL2_ttf/TTF_SetScript
https://github.com/libsdl-org/SDL_ttf/blob/SDL2/SDL_ttf.h#L2281

PR #323, opened 2023-12-27
Updated VisualC project to match harfbuzz sources

pabdulin:
I guess harfbuzz external was updated, but not VisualC msbuild config. Thus attemp to build resulted in `cc` files missing, etc.

I used `dir hb-*.cc` in harfbuzz `src` directory to get updated files list, and after test build link errors indicated that `graph\gsubgpos-context.cc` is also required, so I've added it, after that build was successful.


sezero:
Does the SDL2 branch need similar adjustment?

slouken:
Yes, it does, if we've upgraded harfbuzz on that branch.

sezero:
>  if we've upgraded harfbuzz on that branch.

we did

sezero:
Cherry-picked to SDL2 branch as 25dc937b93f3b41095fabb4ad649fc4baf293e7b

slouken:
Thanks!

Issue #324, opened 2023-12-30
Unaligned access UB in BLIT_GLYPH routines


matoro:
Hi, I identified this bug while testing pygame:  https://bugs.gentoo.org/920956

In the blit glyph routines `BG_64` and `BG_32`, the loops invoke undefined behavior by casting some offset into a larger integer to `u8`.  GCC UBSAN at least catches and warns about this.  These problems are easy to catch on sparc, because unaligned access is completely illegal there and crashes the program with a SIGBUS, but this is not a platform-specific problem:  unaligned access is undefined behavior on all platforms.

Minimized reproducer:

```
$ wget 'https://github.com/pygame/pygame/raw/main/src_py/freesansbold.ttf'
...
$ cat test.c
#include <SDL2/SDL_ttf.h>

int main()
{
        TTF_Init();
        TTF_RenderUTF8_Solid(
                TTF_OpenFont("./freesansbold.ttf", 24),
                "Test",
                (struct SDL_Color){ 255, 255, 255, 255 }
        );
        return 0;
}
$ gcc -ggdb3 -fsanitize=undefined test.c -o test -lSDL2_ttf
$ ./test
/var/tmp/portage/media-libs/sdl2-ttf-2.20.2/work/SDL2_ttf-2.20.2/SDL_ttf.c:919:9: runtime error: load of misaligned address 0x0001272a9817 for type 'const Uint64', which requires 8 byte alignment
0x0001272a9817: note: pointer points here
 00 00 00 00 01  01 01 01 01 01 01 01 01  01 01 01 01 01 00 00 00  00 00 00 00 01 01 01 01  01 01 01
             ^ 
```

And this is the complete backtrace at the point of the crash on sparc:

```
(gdb) frame 0
#0  0xfff80001003039c4 in BG_64 (image=0x10000209848, destination=0x1000021ab50 "", srcskip=5, dstskip=40)
    at /usr/src/debug/media-libs/sdl2-ttf-2.20.2/SDL2_ttf-2.20.2/SDL_ttf.c:919
919             DUFFS_LOOP4(
(gdb) l
914         Uint32        width  = image->width / 8;
915         Uint32        height = image->rows;
916
917         while (height--) {
918             /* *INDENT-OFF* */
919             DUFFS_LOOP4(
920                 *dst++ |= *src++;
921             , width);
922             /* *INDENT-ON* */
923             src = (const Uint64 *)((const Uint8 *)src + srcskip);
(gdb) bt full
[...]

PR #325, opened 2023-12-31
Use memcpy to avoid explicit unaligned load. (See #324)

1bsyl:


src/SDL_ttf.c
```diff
@@ -883,11 +883,13 @@ static SDL_INLINE void BG_64(const TTF_Image *image, Uint8 *destination, Sint32
...
     Uint32        width  = image->width / 8;
     Uint32        height = image->rows;
-
+    Uint64        tmp;
+    
     while (height--) {
         /* *INDENT-OFF* */
         DUFFS_LOOP4(
-            *dst++ |= *src++;
+            memcpy(&tmp, src++, sizeof(tmp));
```
madebr, line 891: Is `memcpy` ok? Or do we want `SDL_memcpy`?

This creates a dependency on libc, or don't we care about this in the satellite libs?

1bsyl: We want memcpy so that compiler can remove it if possible.
I think this code is almost never compiled because there are neon and sse routines also.

PR #328, opened 2024-01-09
Prevent negative draw line and out of bounds.

1bsyl:
can be reproduced with testapp.c, int seed = 1704796537, at iteration ~270


slouken:
Can you cherry-pick -x this to SDL2?

Thanks!

PR #329, opened 2024-01-09
Fixed bug #324 - prevent unaligned load access

1bsyl:
Fixed bug #324 - prevent unaligned load access
use memcpy so we can expect the compiler to optimize when possible


1bsyl:
yes,
using memcpy and -O2 gives the same output (well almost, but optimized)
(.. so I don't know how ubsan achieves to ignore the unaligned load, since it's the same code ...looking at the source maybe..)

If I use *SDL_memcopy*, it adds SDL_memcpy calls.

matoro:
Tested and this indeed does fix the issue.  Also, @slouken here's proof this compiles to identical assembly on x86:  https://godbolt.org/z/aG1qnvn4o  Note there are only differences at -O0, it gets optimized correctly at -O1 or higher.

PR #330, opened 2024-01-10
Fix ASAN reports: runtime error: left shift of XXX by 24 places cannot be represented in type 'int'

1bsyl:
Fix ASAN reports: runtime error: left shift of XXX by 24 places cannot be represented in type 'int'


slouken:
Merged and cherry-picked to SDL2, thanks!

PR #331, opened 2024-01-12
Removed camel case from xcframework target in Xcode

danieltarazona:
Hello. I removed the camel case from xcFramwork to xcframework target in Xcode for consistency between the SDL_* projects. I'm currently working on a bash script to build SDL3 in macOS.


slouken:
Merged, thanks!

PR #336, opened 2024-02-04
Support text from utf8 file and customized window size.

zhk:
- Added argument [-windowsize w,h] to specify the windows size.
- added argument [-utf8txtfile pathName] to support message from utf8 file.
- removed [-utf8|-unicode], since the utf8 or unicode text cannot be input from console.
- fixed [-wrapped] display.

### DIFF
--- a/VisualC/glfont/glfont.vcxproj
+++ b/VisualC/glfont/glfont.vcxproj
@@ -113,6 +113,7 @@
       <PreprocessorDefinitions>WIN32;NDEBUG;_WINDOWS;HAVE_OPENGL;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <RuntimeLibrary>MultiThreadedDLL</RuntimeLibrary>
       <WarningLevel>Level3</WarningLevel>
+      <AdditionalIncludeDirectories>K:\ProjectsMine\SDL_ttf\external\SDL\include</AdditionalIncludeDirectories>
     </ClCompile>
     <ResourceCompile>
       <PreprocessorDefinitions>NDEBUG;%(PreprocessorDefinitions)</PreprocessorDefinitions>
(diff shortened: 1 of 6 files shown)


VisualC/glfont/glfont.vcxproj
```diff
@@ -113,6 +113,7 @@
       <PreprocessorDefinitions>WIN32;NDEBUG;_WINDOWS;HAVE_OPENGL;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <RuntimeLibrary>MultiThreadedDLL</RuntimeLibrary>
       <WarningLevel>Level3</WarningLevel>
+      <AdditionalIncludeDirectories>K:\ProjectsMine\SDL_ttf\external\SDL\include</AdditionalIncludeDirectories>
```
madebr, line 116: Please remove these system-specific lines.
These visual studio projects assume you have the SDL and SDL_ttf repositories checked out next to each other (and built SDL before).
See lines 99-100 (which probably should use `ProjectDir` instead as we do in [SDL3_net](https://github.com/libsdl-org/SDL_net/blob/b2ca02cfd9e0fe5c05bd6b0523c779b35e76772c/VisualC/SDL_net.vcxproj#L86-L101)

examples/showfont.c
```diff
@@ -47,13 +47,13 @@ typedef enum
 
 typedef struct
 {
-    SDL_Texture *caption;
+    SDL_Texture* caption;
```
madebr, line 50: This is wrong. Can you fix the .clang-format?

examples/showfont.c
```diff
@@ -385,34 +475,45 @@ int main(int argc, char *argv[])
...
     case RENDER_UNICODE:
     {
-        Uint16 *unicode_text = SDL_iconv_utf8_ucs2(message);
-        switch (rendermethod) {
+        Uint16* unicode_text = SDL_iconv_utf8_ucs2(message);
+        switch (rendermethod)
+        {
         case TextRenderSolid:
-            if (wrap) {
+            if (wrap)
```
madebr, line 482: This too?
You might need to update the .clang-format to match the style we use

PR #340, opened 2024-02-29
Pass kerning setting through into harfbuzz shaping

Starbuck5:

PR #341, opened 2024-02-29
Pass kerning setting through into harfbuzz shaping (SDL3)

### LINKED ISSUE #339: Enabling/disabling kerning not having any impact on rendering
Hello, one of the pygame-ce members was working on porting SDL_ttf's `TTF_SetFontKerning` function to pygame-ce's font module (https://github.com/pygame-community/pygame-ce/pull/2718), but we noticed that it didn't seem to have any impact on the text, in all the fonts we tried. I tried arial, times new roman, freesansbold, cascadia mono, a couple other things from my system.

Looking into the source, I noticed `font->use_kerning` is only used in non-harfbuzz code paths. I looked into how to enable/disable kerning in harfbuzz, and I found that `hb_shape` can take an array `hb_feature_t`s that adjust their default features. I'll send a PR for this for SDL2 and SDL3. (I wanted to open an issue as well so there's a record in the issues tab, and in case my solution doesn't work for some reason).

Starbuck5:
Fixes #339 

Version of my patch in #340, but against main instead of SDL2. I tested the SDL2 version, I have not tested this version because I don't have working SDL3 environment the way I do with SDL2. But this area of the code doesn't look like it has changed at all between branches.

### DIFF
--- a/src/SDL_ttf.c
+++ b/src/SDL_ttf.c
@@ -3226,7 +3226,14 @@ static int TTF_Size_Internal(TTF_Font *font,
 
     /* Layout the text */
     hb_buffer_add_utf8(hb_buffer, text, -1, 0, -1);
-    hb_shape(font->hb_font, hb_buffer, NULL, 0);
+    
+    hb_feature_t userfeatures[1];
+    userfeatures[0].tag = HB_TAG('k','e','r','n');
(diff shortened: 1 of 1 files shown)


sezero (approved):
CC: @1bsyl, @slouken

PR #342, opened 2024-03-11
Fix a few CMake image/ttf mixups

martinstarkov:
Noticed that the VC release files contain a minor typo in one of the CMake variables, changed all the mention from image to ttf.

Not sure if this PR should be for SDL3 or SDL2 or both but I'm making a PR for both.


madebr:
Thanks for catching these :)

Issue #344, opened 2024-03-16
`SDL_RWops_renamed_SDL_IOStream` with latest SDL3 commit


Semphriss:
This is probably known already, I'm opening this just to keep track of the progress and so that other people know it's a known issue.

Compiling SDL and SDL_ttf from source reports a very long list of errors related to SDL_RWops being renamed to SDL_IOStream.

[Same issue with SDL_img](https://github.com/libsdl-org/SDL_image/issues/436) - I did not check other satellite libraries (SDL_net etc.)

PR #345, opened 2024-03-18
Updated for the SDL3 RW -> IO changes

slouken:
Fixes https://github.com/libsdl-org/SDL_ttf/issues/344


sezero:
s/freesrc/closeio/ in SDL_ttf.c for consistency with other libs?

`*DPIIO` looks funny and intention requires reading twice? (but that's me..)

Other than those, looks OK to me.

P.S.: SDL_rtf needs similar changes.

slouken:
I plan to switch the Open* variants to TTF_OpenFontWithProperties(), so I'll clean that up then.

PR #346, opened 2024-03-20
Harfbuzz can do kerning positioning even if the font hasn't the data 

1bsyl:
Harfbuzz can do kerning positioning even if the font hasn't the data

this fixes is needed after #341 
and should be cherry-picked for SDL2


1bsyl:
for SDL2 https://github.com/libsdl-org/SDL_ttf/commit/2704f4e3318b3c3141235b06867fc8bab4967055

PR #358, opened 2024-05-10
Atlas First Draft

LevoDeLellis:
It's not complete but there's enough to start using it. I'm not happy with the name 'TTF_DestroyAtlas' and I'll need to improve the header documentation. I'm not sure what to do with the example. Using the hashmap in C++ looks nice but it appears no C++ examples are allowed? The C code is close enough. I'd want to improve both.

Another thing I plan on implementing is a way to reserve a series of rectangles so I can draw into them before creating a texture. IIRC GL supports a texture with 8bit alpha channel (GL_Alpha8 and GL_R8 for pixel shaders). Is it possible to use an alpha texture? Right now I use SDL_PIXELFORMAT_ARGB8888 and I only set the alpha bits (`0x00FFFFFF | ((Uint32)*src << 24);`)

If I can get feedback I'll adjust it while implementing the reserve rectangle code and upload a second draft. I'm not expecting this to be merged

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -350,6 +350,8 @@ if(SDL3TTF_SAMPLES)
     add_executable(glfont examples/glfont.c)
     add_executable(showfont examples/showfont.c)
     add_executable(testapp examples/testapp.c)
+    add_executable(AtlasLoadAll examples/AtlasLoadAll.c)
+    #add_executable(Atlas examples/Atlas.cpp)
 
     set(OpenGL_GL_PREFERENCE GLVND)
     find_package(OpenGL)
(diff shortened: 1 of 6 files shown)


LevoDeLellis:
I don't have a windows machine. It should be easy enough to replace #ifdef __WIN32__ with SDL_PLATFORM_WINDOWS but I'm not sure why the compiler doesn't like my macro for variable length array

slouken:
This code compiles with C89, which doesn't have variable length arrays. I haven't looked at the context, but usually those should be SDL_malloc() or SDL_small_alloc(), which falls back to SDL_malloc if the variable size is too big.

LevoDeLellis:
I have more inspiration. I'll send a new pull request tonight.

PR #359, opened 2024-05-12
Second Atlas Draft (Implemenattion done, examples/docs can improve)

LevoDeLellis:
I don't think there's anything else I want to add to the library. I'll want to improve documentation and examples.

I like the C++ version because of the hash table but both are pretty close. The C++ example uses multiple atlas functions while the C version uses the function that loads everything at once. Here's a picture of the C++ version which loads 2 fonts and intentionally didn't load the 'Y' in the larger font

IIRC GL supports a texture with 8bit alpha channel (GL_Alpha8 and GL_R8 for pixel shaders). Is it possible to use an 8bit format for this? At the moment only SDL_PIXELFORMAT_ARGB8888 is supported. I simply use white and set the alpha bits `(0x00FFFFFF | ((Uint32)*src << 24);`)

Let me know what you think. If it's all good I'll write better documentation and may add more to the example(s)
Screenshot_2024-05-11_21-06-49

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -350,6 +350,8 @@ if(SDL3TTF_SAMPLES)
     add_executable(glfont examples/glfont.c)
     add_executable(showfont examples/showfont.c)
     add_executable(testapp examples/testapp.c)
+    add_executable(AtlasLoadAll examples/AtlasLoadAll.c)
+    #add_executable(Atlas examples/Atlas.cpp)
 
     set(OpenGL_GL_PREFERENCE GLVND)
     find_package(OpenGL)
@@ -361,7 +363,7 @@ if(SDL3TTF_SAMPLES)
         target_link_libraries(glfont PRIVATE OpenGL::GL)
     endif()
 
-    foreach(prog glfont showfont testapp)
+    foreach(prog glfont showfont testapp AtlasLoadAll) #Atlas is cpp so it's not supported?
         sdl_add_warning_options(${prog} WARNING_AS_ERROR ${SDL3TTF_WERROR})
         target_link_libraries(${prog} PRIVATE SDL3_ttf::${sdl3_ttf_target_name})
         target_link_libraries(${prog} PRIVATE ${sdl3_target_name})
--- a/examples/Atlas.cpp
+++ b/examples/Atlas.cpp
@@ -1,0 +1,191 @@
+#include <SDL3/SDL_render.h>
+#include <SDL3/SDL_surface.h>
+#include <stdio.h>
+#include <SDL3/SDL.h>
+#include "SDL_ttf_atlas.h"
+#include <unordered_map>
+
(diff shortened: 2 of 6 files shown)


slouken:
I haven't done a full review, but here are some thoughts:
* To simplify the API, maybe return the atlas pixels as an SDL_Surface? That can be easily converted to a texture or consumed by other drawing code.
* I don't see kerning information used anywhere, and that's important for many fonts.
* The baseline should be already available for a given font, it's the point between the ascent and descent.

LevoDeLellis:
I just noticed how hard to read this is when looking at the diff. I'll have the next PR be based on the main tree.

1. Sure, if you were looking at draft1 I removed that function, in draft two I can add a SDL_Surface and remove the width/height/pixel format
2. Correct no kerning. I never heard of anyone using that when using an atlas. Do you strongly want this? If so I'll look into it
3. Is there a way to get expected line height? I've been treating y as the top of the line (instead of baseline). In the example I treated baseline as the ascent of E + some padding for line height. I prefer using y as the top but if that's weird I can remove the baseline param and have y be the baseline

LevoDeLellis:
I found some time to look at kernings. It appears FreeType will scan the file every time you look up a pair and there's no way to get a table. However stb_truetype.h appears to create a table which I think is an excellent choice

It offers two license, mit and public domain
https://github.com/nothings/stb?tab=License-1-ov-file#readme
https://github.com/nothings/stb/blob/master/stb_truetype.h#L804

I'm interested in implementing kerning with some stb code.
I just want to make sure that you'll accept my code before I go out to do this. I can write a text document of the API (funcs and structs) if you want. A quick summary is 

https://github.com/LevoDeLellis/SDL_ttf/blob/main/include/SDL3_ttf/SDL_ttf.h#L2251
1. Replacing various members in TTF_AtlasInfo with SDL_Surface as you requested
2. Better documentation and param names (does canvasSquare make more sense than canvasWidthHeight?)
3. Change TTF_AtlasInitFixed to use a surface
4. Change the ReserveEntry 2 and 3 function to Pair and Triple
5. Add a function to get the kerning table
6. Requested on the discourse is having a function to render instead of having the user copy/paste that code
7. Possibly an extremely simple hash table to make kerning and character lookups faster (would I would need for the render function)

Thoughts? I'm not in a rush to implement this. Here's the drawing code from the example, I like having the Y be top instead of baseline but if it makes more sense for Y to be baseline that's completely fine and a minor edit https://github.com/LevoDeLellis/SDL_ttf/blob/main/examples/AtlasLoadAll.c#L142

Green-Sky:
did we just loose this code?

madebr:
> did we just loose this code?

No, run
```
git fetch https://github.com/libsdl-org/SDL_ttf pull/359/head
git checkout FETCH_HEAD
```

Issue #365, opened 2024-07-05
Let HarfBuzz decide which script, direction and language to use in case not set


Magueija:
Text has rules and settings needed for its rendering depending on its characters. Scripts, text direction and language are settings that should be changed depending on the characters and their context. These change the visual effect on the rendered text.

At the moment SDL_TTF provides us the possibility to set any of these three settings but always defaults to an invalid setting on each. With `hb_buffer_guess_segment_properties()` these could be deducted if not set and it should be the default for rendering text correctly.

See https://github.com/libsdl-org/SDL_ttf/pull/366

PR #366, opened 2024-07-05
Let HarfBuzz decide which script, direction and language to use in case not set

Magueija:
Text has rules and settings needed for its rendering depending on its characters. Scripts, text direction and language are settings that should be changed depending on the characters and their context. These change the visual effect on the rendered text.

At the moment SDL_TTF provides us the possibility to set any of these three settings but always defaults to an invalid setting on each. With `hb_buffer_guess_segment_properties()` these could be deducted if not set and it should be the default for rendering text correctly.

See https://github.com/libsdl-org/SDL_ttf/issues/365


slouken:
Merged, thanks!

PR #378, opened 2024-09-27
remove internals to render with former "solid" mode

1bsyl:
@slouken 
if we remove the solid rendering, then we may want to remove all that parts that  freetype uses to render with mono mode.


slouken:
Sounds good, thanks!

PR #385, opened 2024-09-29
Added support for custom font atlas and text drawing implementations

slouken:
Included are two implementations for SDL_Surface and SDL_Renderer

### DIFF
--- a/include/SDL3_ttf/SDL_ttf.h
+++ b/include/SDL3_ttf/SDL_ttf.h
@@ -1228,6 +1228,9 @@ extern SDL_DECLSPEC SDL_Surface * SDLCALL TTF_RenderGlyph_LCD(TTF_Font *font, Ui
  */
 typedef struct TTF_TextEngine TTF_TextEngine;
 
+/* Internal data for TTF_Text */
+typedef struct TTF_TextData TTF_TextData;
+
 /**
  * Text created with TTF_CreateText()
  *
@@ -1235,6 +1238,7 @@ typedef struct TTF_TextEngine TTF_TextEngine;
  *
  * \sa TTF_CreateText
  * \sa TTF_CreateText_Wrapped
+ * \sa TTF_GetTextProperties
  * \sa TTF_DestroyText
  */
 typedef struct TTF_Text
@@ -1243,8 +1247,11 @@ typedef struct TTF_Text
     int w;                  /**< The width of this text, in pixels, read-only. */
     int h;                  /**< The height of this text, in pixels, read-only. */
     SDL_FColor color;       /**< The color of the text, read-write. You can change this anytime. */
-    TTF_TextEngine *engine; /**< The engine used to create this text, read-only. */
-    void *internal;         /**< The internal representation of this text, read-only */
+
+    int refcount;           /**< Application reference count, used when freeing surface */
+
+    TTF_TextData *internal; /**< Private */
+
 } TTF_Text;
 
 /**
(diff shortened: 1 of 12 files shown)


slouken:
This isn't fully implemented yet, but I wanted to put it up for sanity checking.

slouken:
The surface-based implementation works, and is useful as a proof of concept.

src/SDL_hashtable.h
```diff
@@ -0,0 +1,64 @@
...
+     in a product, an acknowledgment in the product documentation would be
+     appreciated but is not required.
+  2. Altered source versions must be plainly marked as such, and must not be
+     misrepresented as being the original software.
+  3. This notice may not be removed or altered from any source distribution.
+*/
+#ifndef SDL_hashtable_h_
+#define SDL_hashtable_h_
+
+// this is not (currently) a public API. But maybe it should be!
```
icculus, line 24: Do we want to make this public in SDL's API so we don't have to copy it to other projects?

slouken: I was debating that, but it's SO ugly to use. You have to cast every access and insert. I'm thinking that people who really need it can copy it, but I could be convinced either way.

slouken: Most projects will have std::map or something similar available to them.

include/SDL3_ttf/SDL_ttf.h
```diff
@@ -1190,6 +1222,195 @@ extern SDL_DECLSPEC SDL_Surface * SDLCALL TTF_RenderText_LCD_Wrapped(TTF_Font *f
...
+ *
+ * \threadsafety This function should be called on the thread that created the text.
+ *
+ * \since This function is available since SDL_ttf 3.0.0.
+ *
+ * \sa TTF_CreateRendererTextEngine
+ * \sa TTF_CreateText
+ * \sa TTF_CreateText_Wrapped
+ */
+extern SDL_DECLSPEC bool SDLCALL TTF_DrawRendererText(TTF_Text *text, float x, float y);
```
icculus, line 1329: It's a little weird you can get a TTF_Text from TTF_CreateText that won't work with this, if it's from a surface-based (or some other future thing) engine. At a minimum, the docs should say this, but maybe the interface needs to change. If we don't think there will be other engines in the future, we should probably have a TTF_RendererText type so you can't mix between engines.

(But even there, we need to be document "this can't be used with a _different_ SDL_Renderer, even if this TTF_Text came from a renderer-based engine," because the underlying SDL_Texture isn't shared.)

src/SDL_hashtable.h
icculus: Okay, let's leave it alone. I'm glad this got picked up _after_ the new hashtable code arrived, though!

icculus:
Overall, I'm glad SDL_ttf is getting some love for SDL3.

include/SDL3_ttf/SDL_ttf.h
slouken: I was thinking the function naming would make it obvious, but we can definitely be explicit that the TTF_DrawXXXText() functions only work with text returned from TTF_CreateText() given TTF_XXXTextEngine. This is going to come up again in the GPU text engine shortly, and who knows how people will extend this in the future.

src/SDL_hashtable.h
slouken: Yep, me too. :)

include/SDL3_ttf/SDL_ttf.h
icculus: Maybe just a generic TTF_DrawText() and SDL_ttf knows what engine is supposed to be used and where the text is going? And then we don't have to add a new TTF_Draw*Text function for each new implementation?

slouken: No, because the drawing parameters are different depending on the underlying use case. The surface one takes a destination surface parameter, the renderer one has an implicit renderer but uses float coordinates, the GPU one ... who knows?

I went ahead and documented this better though, thanks!

PR #388, opened 2024-10-01
TTF_Draw* functions should return true on success

captain0xff:
The title

Also, updated the .gitignore to ignore clangd .cache folder.


slouken:
Merged, thanks!

PR #394, opened 2024-10-03
testapp: allow to render text with SDL_TextEngine (key 'x')

1bsyl:
@slouken 
this allow to use testapp with textengine and compare none/surface/renderer

allow, it seems "TE renderer" asserts with: 
```
Assertion failure at CreateDrawSequence (//SDL_ttf_git/src/SDL_renderer_textengine.c:540), triggered 1 time:
  'num_ops > 0'

```
while "TE surface" doesn't assert

### DIFF
--- a/examples/testapp.c
+++ b/examples/testapp.c
@@ -33,6 +33,19 @@
 
 static const int g_force_no_SDF = 0; /* make random fuzzer faster by disabling SDF rendering */
 
+//#define HAVE_LCD
+#define HAVE_SDF
+
+#define HAVE_SET_FONT_SIZE_FUNCTION
+// Need patch to set size dynamically
+// https://bugzilla.libsdl.org/show_bug.cgi?id=2487
+
+#define HAVE_WRAP_ALIGN
+
+
+
+
+
 #include <stdio.h>
 #include <stdlib.h>
 #include <time.h>
@@ -100,6 +113,7 @@ static void help(void)
     SDL_Log("d   : display normal texture, no screen update, stream texture ");
     SDL_Log("r   : start/stop random test");
     SDL_Log("m   : render mode Blended/Shaded");
+    SDL_Log("x   : text engine None/Surface/Renderer");
     SDL_Log("n   : change direction");
     SDL_Log("9/0 : -/+ alpha color fg");
     SDL_Log("7/8 : -/+ alpha color bg (Shaded only)");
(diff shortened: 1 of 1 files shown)


examples/testapp.c
```diff
@@ -1036,30 +1118,68 @@ int main(void)
...
-                d.h = (float)dstrect.h;
-                SDL_RenderTexture(renderer, text_texture, NULL, &d);
-             }
-             SDL_RenderPresent(renderer);
+              int w, h;
+              float x, y;
+              TTF_GetTextSize(text_obj, &w, &h);
+
+              x = windoww/2.0 - w/2.0;
+              y = windowh/2.0 - h/2.0;
```
sezero, line 1164: `2.0f`

1bsyl: done:)

slouken:
This seems fine, but can you give me a standalone repro for the num_ops assertion?

1bsyl:
@slouken 
it seems this happens with the first rendering of testapp :
fonts/FletcherGothicFLF.ttf
and this is string in  test_strings[0] 

just do ./testapp and press x x and it asserts.
you need the fonts file also in "./fonts/FletcherGothicFLF.ttf"

slouken:
I'm not reproducing this with the current head. Has it been fixed?

1bsyl:
@slouken 
this is still present in current HEAD.
renderer_surface would also assert, but the is no "SDL_assert(num_ops > 0);"

after, Render_Line_TextEngine num_ops is still 0


Render_Line_TextEngine:
pos_len is 4. The 4 glyphs are found, but  widths and rows are 0. so no `op_index++
`glyph->sz_width == glyph->sz_rows == 0`
and last `glyph_width == glyph_rows == 0`


the string is of course wrong because testapp is meant to break rendering with corner case.
I compile with SDL_ttf:
```
CMAKE_FLAGS+=" -DSDLTTF_VENDORED=ON "
        CMAKE_FLAGS+=" -DSDLTTF_HARFBUZZ=ON "
```

slouken:
I'm not seeing this here at all. Maybe my version of the font is different? Can you create an issue and provide a modification to showfont and attach a link to the font you're using so I can repro in a debugger?

1bsyl:
`./showfont -textengine renderer fonts/FletcherGothicFLF.ttf  20 "纸"`

1bsyl:
then font I have [FletcherGothicFLF.ttf.txt](https://github.com/user-attachments/files/17261612/FletcherGothicFLF.ttf.txt)

1bsyl:
@slouken here it is

PR #396, opened 2024-10-03
Compilation without HarfBuzz fixed

brainstream:
`TTF_Font::hb_direction` and `TTF_Font::hb_script` are under conditional compilation, but they are assigned without it.


slouken:
Merged, thanks!

PR #400, opened 2024-10-04
Fix shadowed variables

1bsyl:


sezero:
Pushed 599254844107e19df52842f4e9f75cd2895a4815 to clean-up/simplify.

PR #412, opened 2024-10-17
Added GPUTextEngine

captain0xff:
This PR extends the new text rendering api to support the gpu api. It's still not ready for merging but the reason I am creating the PR is so that the API can be reviewed. 

Things that I still need to do:-
- [x] The API needs to be documented (I will do this after the public API is finalized)
- [x] The math_3d.h header should be removed in favor of custom math
- [x] Shaders should be added for all the backends along with a proper workflow to load and build them
- [x] The textgputext example should be improved to show the full capabilities of the api

I will need some help with the shaders because I don't have much idea about the workflow needed for the Metal and Direct3D backends.


captain0xff:
Looks like stuff broke after pulling the latest changes. I am working on fixing them.

captain0xff:
Fixed.
The SDL_gpu_textengine.c is mostly similar to the SDL_renderer_textengine.c. I have just adjusted most stuff for the gpu api.
I have added TTF_GetGPUTextDrawData to return the user the needed geometry data to draw the text and removed the Draw function.  The UpdateGPUTexture method is stolen from the renderer gpu backend and can/should be improved. Apart from this I have updated the CreateDrawSequence to cache the positions along with the texcoords and indices.

examples/testgputext.c
```diff
@@ -0,0 +1,393 @@
...
+typedef struct GeometryData {
+	Vertex *vertices;
+	int vertex_count;
+	int *indices;
+	int index_count;
+} GeometryData;
+
+
+void check_error_bool(const bool res) {
+	if (!res)
```
slouken, line 50: Please use SDL coding style, e.g. 4 spaces, braces on the same line, etc.

slouken:
In general, this looks good so far!

examples/testgputext.c
captain0xff: Is there some script or tool I can use to format it automatically?

slouken: Yep, you can use clang-format.

examples/SDL_math3d.h
```diff
@@ -0,0 +1,261 @@
...
+
+#include <SDL3/SDL_stdinc.h>
+
+// Set up for C function definitions, even when using C++
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#ifndef M_PI
+#define M_PI 3.14159265358979323846
```
slouken, line 17: You should be able to use SDL_PI_D or SDL_PI_F

captain0xff:
I have removed the hardcoded shader stuff and took the script from src/render/gpu/shader to build my shaders. I have only tested the workflow for spirv and vulkan since I only have access to a linux machine atm. I will appreciate if someone can help generate the shaders for other platforms or even provide some input about them. The renderer gpu backend doesn't seem to have any hlsl or msl shader present so I assume that the script generates them from spirv provided you have spirv-cross in your path.
Next, I will clean up and improve the example. It may take some time depending on how quickly I can grasp the math behind perspective and lookAt functions and implement them. I can copy them from somewhere but I like to understand the math behind stuff I do. Also, that can cause licensing issues. 
After, that hopefully that the PR will be ready for a final review once I address the CI.

captain0xff:
I have improved the example and also formatted it

captain0xff:
I have fixed the CI.
I am marking the PR ready. It still needs the shaders for the other platforms but I don't have access to a mac and windows machine.

examples/testgputext.c
```diff
@@ -0,0 +1,381 @@
...
+    Vertex *vertices;
+    int vertex_count;
+    int *indices;
+    int index_count;
+} GeometryData;
+
+void check_error_bool(const bool res)
+{
+    if (!res)
+        SDL_LogError(SDL_LOG_CATEGORY_APPLICATION, "%s", SDL_GetError());
```
madebr, line 52: Better safe then sorry. ;)
There are other instances of this.

```suggestion
    if (!res) {
        SDL_LogError(SDL_LOG_CATEGORY_APPLICATION, "%s", SDL_GetError());
    }
```

captain0xff: :(

madebr:
> I have fixed the CI. I am marking the PR ready. It still needs the shaders for the other platforms but I don't have access to a mac and windows machine.

At https://github.com/libsdl-org/SDL_gpu_shadercross/pull/27, we're working on a cli util that can generate shaders for various formats. There are binaries available for msvc, linux and macos on the GitHub actions page.

captain0xff:
With the help of a person from the FNA discord I have added the d3d12 and metal shaders. Though note that they are added from linux and are untested.
The d3d11 shaders can also be added from linux but wine is needed. I use debian sid and at it's current state there are some issues with wine.
@madebr I have seen you experimenting with wine in the past so can you do it? What you need to do is compile this thing https://github.com/mozilla/fxc2 (check the makefile) and add a scipt named `fxc` that runs the compiled program with wine somewhere in your path. Then you can just run the ./build-shader.sh script.

captain0xff:
After the d3d11 shaders are added I will hook these  shaders with the examples.

madebr:
I think you don't need to commit all files in the shaders directory.
See https://github.com/libsdl-org/SDL/tree/a10578acbdd86924574c9d5863b698b90a4ae212/src/render/gpu/shaders

thatcosmonaut:
Now that we have a CLI for shadercross we should be building shaders with that instead, no need for another shader frontend dependency.

captain0xff:
> I don't think you need to commit all files in the shaders directory. See https://github.com/libsdl-org/SDL/tree/a10578acbdd86924574c9d5863b698b90a4ae212/src/render/gpu/shaders, there are no `.spv` files there.

Yeah, pushed them by mistake. I will remove those files.

examples/testgputext.c
```diff
@@ -222,11 +238,11 @@ int main(int argc, char *argv[])
...
     context.window = check_error_ptr(SDL_CreateWindow("GPU text test", 800, 600, 0));
 
-    context.device = check_error_ptr(SDL_CreateGPUDevice(SDL_GPU_SHADERFORMAT_SPIRV, true, NULL));
+    context.device = check_error_ptr(SDL_CreateGPUDevice(SUPPORTED_SHADER_FORMATS, true, NULL));
     check_error_bool(SDL_ClaimWindowForGPUDevice(context.device, context.window));
 
-    SDL_GPUShader *vertex_shader = load_shader(context.device, shader_vert_spv, shader_vert_spv_len, true, 0, 1, 0, 0);
-    SDL_GPUShader *fragment_shader = load_shader(context.device, shader_frag_spv, shader_frag_spv_len, false, 1, 0, 0, 0);
+    SDL_GPUShader *vertex_shader = load_shader(context.device, true, 0, 1, 0, 0);
+    SDL_GPUShader *fragment_shader = load_shader(context.device, false, 1, 0, 0, 0);
```
madebr, line 245: I think it's worth to check the shader pointer, and log on failure.

captain0xff:
Note, the example seems to leak memory. But it's probably due to https://github.com/libsdl-org/SDL_ttf/issues/425 and so not caused by me.

madebr:
> Note, the example seems to leak memory. But it's probably due to #425 and so not caused by me.

You can verify by linking to SDL3_test, and calling `SDLTest_TrackAllocations(void);` at the very beginning, and calling `SDLTest_LogAllocations(void)` at the very end.

captain0xff:
Well it's show allocations and those are increasing.

captain0xff:
Just to make sure, this PR isn't waiting for something from my end right?

examples/testgputext/fonts/license.txt
```diff
@@ -0,0 +1,37 @@
+License
```
slouken, line 1: Please don't include any fonts in the SDL_ttf project. There are plenty out there already for people to use.

slouken (commented):
This looks good to me!

examples/testgputext/fonts/license.txt
captain0xff: Copied from discord.
> I actually decided to ship the font because if the font isn't monospaced the example will look quite weird and also different fonts will change the size of the text drastically due to the 3D rotation, it will be very hard to mitigate that only by changing the point size.

examples/testgputext.c
```diff
@@ -0,0 +1,401 @@
...
+                    .color_blend_op = SDL_GPU_BLENDOP_ADD,
+                    .color_write_mask = 0xF,
+                    .src_alpha_blendfactor = SDL_GPU_BLENDFACTOR_SRC_ALPHA,
+                    .dst_alpha_blendfactor = SDL_GPU_BLENDFACTOR_DST_ALPHA,
+                    .src_color_blendfactor = SDL_GPU_BLENDFACTOR_SRC_ALPHA,
+                    .dst_color_blendfactor = SDL_GPU_BLENDFACTOR_ONE_MINUS_SRC_ALPHA
+                }
+            }},
+            .has_depth_stencil_target = false,
+            .depth_stencil_format = SDL_GPU_TEXTUREFORMAT_INVALID /* Neex to set this to avoid missing initializer for field error */
```
MirkoCovizzi, line 264:
```suggestion
            .depth_stencil_format = SDL_GPU_TEXTUREFORMAT_INVALID /* Need to set this to avoid missing initializer for field error */
```

MirkoCovizzi (commented):
Thanks for this contribution! I'm implementing fonts in my project and this is a very good learning material!

Just a minor typo

examples/testgputext.c
```diff
@@ -0,0 +1,401 @@
...
+}
+
+void free_context(Context *context)
+{
+    SDL_ReleaseGPUTransferBuffer(context->device, context->transfer_buffer);
+    SDL_ReleaseGPUSampler(context->device, context->sampler);
+    SDL_ReleaseGPUBuffer(context->device, context->vertex_buffer);
+    SDL_ReleaseGPUBuffer(context->device, context->index_buffer);
+    SDL_ReleaseGPUGraphicsPipeline(context->device, context->pipeline);
+    SDL_DestroyGPUDevice(context->device);
```
MirkoCovizzi, line 225: This is missing `SDL_ReleaseWindowFromGPUDevice(context->device, context->window);`

examples/testgputext.c
```diff
@@ -0,0 +1,401 @@
...
+        createinfo.entrypoint = is_vertex ? "VSMain" : "PSMain";
+    } else if (format & SDL_GPU_SHADERFORMAT_DXIL) {
+        createinfo.format = SDL_GPU_SHADERFORMAT_DXIL;
+        createinfo.code = is_vertex ? shader_vert_sm60_dxil : shader_frag_sm60_dxil;
+        createinfo.code_size = is_vertex ? SDL_arraysize(shader_vert_sm60_dxil) : SDL_arraysize(shader_frag_sm60_dxil);
+        createinfo.entrypoint = is_vertex ? "VSMain" : "PSMain";
+    } else if (format & SDL_GPU_SHADERFORMAT_METALLIB) {
+        createinfo.format = SDL_GPU_SHADERFORMAT_METALLIB;
+        createinfo.code = is_vertex ? shader_vert_metal : shader_frag_metal;
+        createinfo.code_size = is_vertex ? shader_frag_metal_len : shader_frag_metal_len;
```
MirkoCovizzi, line 98:
```suggestion
        createinfo.code_size = is_vertex ? shader_vert_metal_len : shader_frag_metal_len;
```

captain0xff:
Done. Thanks for the review.

examples/testgputext/SDL_math3d.h
```diff
@@ -0,0 +1,309 @@
...
+{
+    return SDL_Matrix4X4(
+         1,  0,  0,  offset.x,
+         0,  1,  0,  offset.y,
+         0,  0,  1,  offset.z,
+         0,  0,  0,         1
+    );
+}
+
+static inline SDL_Mat4X4 SDL_MatrixRotationX(float angle)
```
MirkoCovizzi, line 223: I believe this should be instead:

```suggestion
static inline SDL_Mat4X4 SDL_MatrixRotationX(float angle)
{
    float cos = SDL_cos(angle);
    float sin = SDL_sin(angle);

    return SDL_Matrix4X4(
         1,  0,    0,  0,
         0, cos, -sin,  0,
         0, sin,  cos,  0,
         0,  0,    0,  1
    );
}
```
At the moment this function is a copy of `SDL_MatrixRotationY`.

captain0xff:
https://github.com/libsdl-org/SDL/issues/11398 was closed as not planned. So, this PR is again ready I hope?

slouken:
@captain0xff, can you rebase and resolve conflicts?

slouken:
Actually, never mind, I can squash this as a single commit.

captain0xff:
Yes, that will be better. Most of the commit messages are not very sensible.

slouken:
Merged, thanks!

Issue #425, opened 2024-10-25
Memory leak when using TTF_CreateText


bmrhoads:
I'm having an issue with this and I've isolated it to a simple project with a main like so 

```
#include <SDL3/SDL.h>
#include <SDL3_ttf/SDL_ttf.h>

int main(int argc, char *argv[]) {
  int sdlVer = SDL_GetVersion();
  SDL_Window *window = 0;
  SDL_Renderer *renderer = 0;
  SDL_CreateWindowAndRenderer("Test", 900, 500, SDL_WINDOW_RESIZABLE, &window, &renderer);

  TTF_Init();
  TTF_TextEngine *engine = TTF_CreateRendererTextEngine(renderer);
  TTF_Font *font = TTF_OpenFont("dogica.ttf", 12);

  while (1) {
    TTF_Text *text = TTF_CreateText(engine, font, "Testing", sizeof("Testing"));
    SDL_RenderClear(renderer);
    TTF_DrawRendererText(text, 100, 100);
    SDL_RenderPresent(renderer);
    TTF_DestroyText(text);
  }

  TTF_CloseFont(font);
  TTF_DestroyRendererTextEngine(engine);
  TTF_Quit();
  SDL_DestroyWindow(window);
  SDL_Quit();

  return 0;
}
```

This process of creating the TTF_Text pointer and then destroying it seems to still leave something allocated because the memory grows at ~1MB a second. Maybe I'm misunderstanding something about how CreateText is meant to be used, I'm not sure.

This is on the newest build of the main branch

slouken:
I believe that should work. I'll double check.

captain0xff:
It may be unrelated but one thing I can spot is that sizeof("Testing") will include the null character, you probably don't want that.

bmrhoads:
Oh you're right, I should use strlen since CreateText appends the null terminator. I did notice that DestroyEngineText is not firing since these are new Text instances and needs_layout_update is always false. Could it be that the engine->user_data is not being destroyed when the text is destroyed?

slouken:
Fixed!

PR #430, opened 2024-11-13
Fix memory leak when destroying text

MirkoCovizzi:
The engine's `DestroyText` function is
never called, as well as the frees for some
internal components.
This fixes memory leaks reported by some users.

### DIFF
--- a/src/SDL_ttf.c
+++ b/src/SDL_ttf.c
@@ -4863,6 +4863,18 @@ void TTF_DestroyText(TTF_Text *text)
         return;
     }
 
+    DestroyEngineText(text);
+
+    if (text->internal->ops) {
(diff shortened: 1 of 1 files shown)


MirkoCovizzi:
Fixes: https://github.com/libsdl-org/SDL_ttf/pull/412#issuecomment-2448128287 (tested)
Should fix also: https://github.com/libsdl-org/SDL_ttf/issues/425

MirkoCovizzi:
@captain0xff Please let me know if this works for you

captain0xff:
Yes, this fixes the leak on my end.

slouken:
Merged, thanks!

bmrhoads:
Yeah, it's working great thank you

PR #442, opened 2024-12-21
Prevent infinite recursion in TTF_RenderGlyph_Solid

### LINKED ISSUE #441: Infinite recursion in TTF_RenderGlyph_Solid on Linux when using TrueType rendering
When calling _TTF_RenderGlyph_Solid_ passing _0 (zero)_ as the codepoint to render the problem occurs, deep in the call stack at _TTF_Size_Internal_.

```
// Call stack
TTF_RenderGlyph_Solid
TTF_RenderText_Solid
TTF_Render_Internal
TTF_Size_Internal
```

It appers that this problem was introduced on commit [c96f8fb1](https://github.com/libsdl-org/SDL_ttf/commit/c96f8fb1) by this change

changes_on_c96f8fb1

here **text** is a 4 byte array with the first element _0 (zero)_ and the others _uninitialized_ as returned by _SDL_UCS4ToUtf8_ in _TTF_RenderGlyph_Solid_.

When _SDL_StepUTF8_ is called, it checks if the current codepoint is _0 (zero)_ and returns without decrement the length, causing the issue.

Here is a minimal example that triggers this behavior

```cpp
#include <iostream>

#include <SDL3/SDL.h>
#include <SDL3_ttf/SDL_ttf.h>

int
main()
{
    if (!TTF_Init()) {
        std::cout << "TTF_Init: " << SDL_GetError() << "\n";

        return 1;
    }

    TTF_Font *ttf_font = TTF_OpenFont("Bm437_IBM_VGA_9x16.otb", 9);

    if (ttf_font == nullptr) {
        std::cout << "Error: Font not found.\n";

        return 2;
    }

    constexpr SDL_Color kRenderColor{0xff, 0xff, 0xff, 0xff};

    SDL_Surface *glyph = TTF_RenderGlyph_Solid(ttf_font, 0, kRenderColor);

    TTF_Quit();

    return 0;
}
```
 and the project [sdl3_font_rendering.zip](https://github.com/user-attachments/files/18219138/sdl3_font_rendering.zip) that I used for the tests.

croscato:
Prevent infinite recursion in TTF_RenderGlyph_Solid on Linux when using TrueType rendering

Fixes #441

### DIFF
--- a/src/SDL_ttf.c
+++ b/src/SDL_ttf.c
@@ -3111,6 +3111,10 @@ static bool TTF_Size_Internal(TTF_Font *font, const char *text, size_t length, i
         if (c == UNICODE_BOM_NATIVE || c == UNICODE_BOM_SWAPPED) {
             continue;
         }
+
+        if (c == 0 && length > 0) {
+            --length;
+        }
 #endif
         if (!Find_GlyphByIndex(font, idx, 0, 0, 0, 0, 0, 0, &glyph, NULL)) {
             goto failure;


slouken:
Merged, thanks!

Issue #439, opened 2024-12-16
`testgputext` doesn't correctly apply vertex offset


Beyley:
https://github.com/libsdl-org/SDL_ttf/blob/0ad0c2518be948300ed1e133741932f1ba2a4e7b/examples/testgputext.c#L210

That second to last parameter should be the vertex offset (which you calculate by tallying `seq->num_vertices` as you go).

As per 

https://github.com/libsdl-org/SDL_ttf/blob/0ad0c2518be948300ed1e133741932f1ba2a4e7b/include/SDL3_ttf/SDL_ttf.h#L1661

The indices that `testgputext` copies onto the GPU for sequences are indexed into the per sequence `vertices` array, not a global tally of total vertices, so you need to specify the vertex offset in the draw call to correct for that.

slouken:
@captain0xff, can you take a look?

Beyley:
closing as fixed

Issue #438, opened 2024-12-16
`TTF_GPUAtlasDrawSequence.xy/uv` un-ergonomic stride behaviour


Beyley:
```c
float *xy;                              /**< Vertex positions */
int xy_stride;                          /**< Byte size to move from one element to the next element */

...

sequence->xy_stride = sizeof(float) * 2;
sequence->uv_stride = sizeof(float) * 2;

...

const float *xy = (float *)((Uint8 *)sequence->xy + i * sequence->xy_stride);
```

At the moment to use the `xy`/`uv` fields (which are both `float *`), you have to cast it to a `Uint8 *`, then apply the stride (which is in bytes), then cast back to a `float *` to be able to read the value back out.

Given the strides are always constant and always defined as a multiple of `sizeof(float)`, it would make sense for these strides to be changed to be based on values, rather than bytes, as the current API is very error prone, given you have to hot-potato the pointer type.

pinging @captain0xff since you wrote this initially

slouken:
Strides are generally calculated in bytes, since the values might be pointing into structures that are not aligned with the size of a float.

slouken:
@captain0xff, is there a reason this is variable? Should we just document that the xy and uv coordinates are pairs?

captain0xff:
You are right this doesn't need to be a variable at all. This is just wasting memory unnecessarily. We can just document that they are pairs or provide a macro for the strides or do both.

captain0xff:
I will fix this along with the other two issues.

captain0xff:
I have removed the stride parameters and added to the docs that xy/uv are contiguous arrays consisting of a pair of floats. With the current implementation I don't see a need to do add the strides as macros. @slouken what's your opinion on this?

slouken:
> With the current implementation I don't see a need to add the strides as macros. @slouken what's your opinion on this?

Agreed.

captain0xff:
I am making the xy/uv an array of SDL_FPoint because it feels the something like the following is equally error prone and weird
```c
vert.pos = (Vec3){ sequence->xy[2 * i], sequence->xy[2 * i + 1], 0.0f };
// Or
const float *xy = sequence->xy + 2 * i;
vert.pos = (Vec3){ xy[0], xy[1], 0.0f };
```

Beyley:
closing as fixed

Issue #437, opened 2024-12-16
[Feature Request] Allow specifying the winding order of GPU indices


Beyley:
https://github.com/libsdl-org/SDL_ttf/blob/0ad0c2518be948300ed1e133741932f1ba2a4e7b/src/SDL_gpu_textengine.c#L667

The default order that SDL_ttf outputs is opposite to the rest of my pipeline, so i have to do a post-processing step over the returned indices for it to render correctly, which is quite wasteful, it would be nice if i could specify the winding order for SDL_ttf's GPU text engine

slouken:
@captain0xff, can you take a look?

captain0xff:
I have fixed the other two issues and will create a PR after addressing this issue. I am just confused regarding whether the option should be a setter/getter or a hint. Though none of the satellite libs seems to use hints. @slouken is there a reason for this? And is there a policy regarding whether something should be a hint or api call?

slouken:
Typically we don't set global options in the library, since it might be used by two different pieces of code with different requirements. I think that's unlikely in this case though.

An option when creating the text engine seems like a reasonable approach to me...?

captain0xff:
> An option when creating the text engine seems like a reasonable approach to me...?

This is a possible way but probably 95% of the users won't use it or even know its significance. That's why it feels a bit odd to force them to set an extra parameter while creating the text engine.
One thing we can do is have TTF_Set/GetGPUTextEngineVertexWindingOrder that takes the text engine as a parameter. And we can leave the default winding order as clockwise. This way the option isn't global, beginners won't be bothered by it and people who want to do culling can set it as per their choice.

slouken:
Sounds good.

PR #443, opened 2024-12-22
GPU text engine api improvements

captain0xff:
Fixes: 
- [x] #439 
- [x] #438
- [ ] #437 (waiting for https://github.com/libsdl-org/SDL_ttf/issues/437#issuecomment-2558216191)

### DIFF
--- a/src/SDL_gpu_textengine.c
+++ b/src/SDL_gpu_textengine.c
@@ -18,6 +18,7 @@
      misrepresented as being the original software.
   3. This notice may not be removed or altered from any source distribution.
 */
+#include <SDL3/SDL.h>
 #include <SDL3_ttf/SDL_textengine.h>
 
 #include "SDL_hashtable.h"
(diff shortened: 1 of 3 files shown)


captain0xff:
Let me know if there are any concerns regarding the SDL_FPoint* to float* casts. I have added an assert for safety. But in case this is a bad practice I will remove those.

src/SDL_gpu_textengine.c
```diff
@@ -588,6 +588,7 @@ static AtlasDrawSequence *CreateDrawSequence(TTF_DrawOperation *ops, int num_ops
     }
 
     SDL_assert(num_ops > 0);
+    SDL_assert(sizeof(SDL_FPoint) == 2 * sizeof(float));
```
slouken, line 591: This should be `SDL_COMPILE_TIME_ASSERT()`

PR #447, opened 2024-12-28
Fixed a typo in the documentation for TTF_OpenFontWithProperties and added TTF_FontStyleFlags

edwardgushchin:
The TTF_SetFontStyle method does not accept any value other than TTF_STYLE_NORMAL, TTF_STYLE_BOLD, TTF_STYLE_ITALIC, TTF_STYLE_UNDERLINE, TTF_STYLE_STRIKETHROUGH and never will in the future.

The TTF_GetFontStyle method will never return a value other than these either.

That's why, I think, it will be more convenient and intuitive to operate with a flag and will help to avoid errors when a value that does not correspond to the value of constants is passed to the method.

### DIFF
--- a/include/SDL3_ttf/SDL_ttf.h
+++ b/include/SDL3_ttf/SDL_ttf.h
@@ -354,6 +354,9 @@ extern SDL_DECLSPEC bool SDLCALL TTF_GetFontDPI(TTF_Font *font, int *hdpi, int *
 /**
  * Font style flags
  */
+
+typedef Sint32 TTF_FontStyleFlags;
+
 #define TTF_STYLE_NORMAL        0x00
 #define TTF_STYLE_BOLD          0x01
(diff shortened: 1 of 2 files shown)


include/SDL3_ttf/SDL_ttf.h
```diff
@@ -354,6 +354,9 @@ extern SDL_DECLSPEC bool SDLCALL TTF_GetFontDPI(TTF_Font *font, int *hdpi, int *
 /**
  * Font style flags
  */
+
+typedef Sint32 TTF_FontStyleFlags;
```
slouken, line 358: If we're going to typedef this, it should be Uint32 and have more complete documentation, e.g.:
https://github.com/libsdl-org/SDL/blob/ec8a780ad47b397dc791bec2fa8bd307c9dbbc2c/include/SDL3/SDL_init.h#L64-L87

PR #451, opened 2024-12-28
Added TTF_FontStyleFlags and TTF_HintingFlags

edwardgushchin:
Added TTF_FontStyleFlags and TTF_HintingFlags.

### DIFF
--- a/include/SDL3_ttf/SDL_ttf.h
+++ b/include/SDL3_ttf/SDL_ttf.h
@@ -460,7 +460,7 @@ extern SDL_DECLSPEC int SDLCALL TTF_GetFontOutline(const TTF_Font *font);
 /**
  * Hinting flags for TTF (TrueType Fonts)
  *
- * These flags specify the level of hinting to be applied to the font rendering.
+ * This enum specifies the level of hinting to be applied to the font rendering.
  * The hinting level determines how much the font's outlines are adjusted for better alignment on the pixel grid.
  *
  * \since This enum is available since SDL_ttf 3.0.0.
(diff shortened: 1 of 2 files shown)


src/SDL_ttf.c
```diff
@@ -247,7 +247,7 @@ struct TTF_Font {
     int lineskip;
 
     // The font style
-    int style;
+    Uint32 style;
```
slouken, line 250: Shouldn't this be TTF_FontStyleFlags? And the same for prev_style below?

include/SDL3_ttf/SDL_ttf.h
```diff
@@ -447,13 +458,23 @@ extern SDL_DECLSPEC bool SDLCALL TTF_SetFontOutline(TTF_Font *font, int outline)
...
+ *
+ * \sa TTF_SetFontHinting
+ * \sa TTF_GetFontHinting
  */
-#define TTF_HINTING_NORMAL          0
-#define TTF_HINTING_LIGHT           1
-#define TTF_HINTING_MONO            2
-#define TTF_HINTING_NONE            3
-#define TTF_HINTING_LIGHT_SUBPIXEL  4
+typedef Uint32 TTF_HintingFlags;
```
slouken, line 471: Since these values are exclusive, maybe this should be an enum?

edwardgushchin: Yeah, you're right.

src/SDL_ttf.c
edwardgushchin: I've been thinking about it. Wasn't sure it wouldn't break anything.

include/SDL3_ttf/SDL_ttf.h
```diff
@@ -447,13 +458,24 @@ extern SDL_DECLSPEC bool SDLCALL TTF_SetFontOutline(TTF_Font *font, int outline)
 extern SDL_DECLSPEC int SDLCALL TTF_GetFontOutline(const TTF_Font *font);
 
 /**
- * Hinting flags
+ * Hinting flags for TTF (TrueType Fonts)
+ *
+ * These flags specify the level of hinting to be applied to the font rendering.
```
slouken, line 463:
```suggestion
 * This enum specifies the level of hinting to be applied to the font rendering.
```

slouken:
Merged, thanks!

Issue #452, opened 2024-12-30
MacOS dpi scaling & filtering issues with GPUTextEngine


bungoboingo:
Few issues with the GPUTextEngine example on MacOS (ARM)

https://github.com/user-attachments/assets/c2e0fb16-a310-4c4e-8180-8ee1a1b837e0

**1) Glyph atlas doesn't account for DPI scaling, leading to blurry text due to upsampling**
It would be great to be able to pass DPI scaling into `CreateGPUTextEngine`, and have some way to update it if DPI scaling changes (or just have this be automatic change when the event is propagated)

**2) Filtering seems a bit off, this may (?) be fixed by accounting for DPI scaling however**

I believe also that in the GPUEngine example the format should be `SDL_GPU_SHADERFORMAT_MSL` w/ entrypoint `"main0"` not `SDL_GPU_SHADERFORMAT_METALLIB` / `vs_main`, I had to switch it to be able for SDL to make the vertex shader at all.

slouken:
@captain0xff, can you take a look?

captain0xff:
@TheSpydog has fixed 

> Few issues with the GPUTextEngine example on MacOS (ARM)
> macos_txt_gpu_example.mov
> 
> **1) Glyph atlas doesn't account for DPI scaling, leading to blurry text due to upsampling** It would be great to be able to pass DPI scaling into `CreateGPUTextEngine`, and have some way to update it if DPI scaling changes (or just have this be automatic change when the event is propagated)
> 
> **2) Filtering seems a bit off, this may (?) be fixed by accounting for DPI scaling however**
> 
> I believe also that in the GPUEngine example the format should be `SDL_GPU_SHADERFORMAT_MSL` w/ entrypoint `"main0"` not `SDL_GPU_SHADERFORMAT_METALLIB` / `vs_main`, I had to switch it to be able for SDL to make the vertex shader at all.

@TheSpydog has fixed the second issue. And regarding the first issue I don't have a good solution. The problem is that the glyph's size on the text atlas are only dependent on the point size of the font you use to draw the text with the engine. So, this is not something that the engine can control. There will be always blurring due to upscaling if the user doesn't use a sufficiently big point size. The solution for this is probably having support for SDF. There are some plans for it but no one is working on it atm. I will be happy to review if someone comes up with a PR or even a working prototype for this.

bungoboingo:
First issue is solved, thanks so much! 😃 

[...]

PR #455, opened 2025-01-01
GPU Metal fixes

TheSpydog:
The Metal shaders provided are actually MSL, just in a byte array format, so I made some fixes to properly load the shaders.

Additionally, the atlas's texture contents are undefined outside of the regions that have been explicitly uploaded, so we should explicitly clear the atlas texture upfront. This fixes the pink line issue mentioned here: https://github.com/libsdl-org/SDL_ttf/issues/452

### DIFF
--- a/examples/testgputext.c
+++ b/examples/testgputext.c
@@ -11,7 +11,7 @@
 
 #define MAX_VERTEX_COUNT 4000
 #define MAX_INDEX_COUNT  6000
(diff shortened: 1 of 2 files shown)


slouken:
Merged, thanks!

bungoboingo:
Thanks so much for the quick fix! :-)

PR #458, opened 2025-01-02
Use freetype library if it already exists

Ravbug:
If one is building freetype without making it a subdirectory of SDL_ttf, telling SDL_ttf to use it is rather nasty:
```cmake
add_subdirectory(freetype EXCLUDE_FROM_ALL)
if(NOT TARGET Freetype::Freetype)              # freetype doesn't make this on its own but SDL_ttf is hardcoded to link it
    add_library(Freetype::Freetype ALIAS freetype)
endif()
set(SDLTTF_VENDORED OFF CACHE INTERNAL "")     # unintuitive - aren't we vendoring it?
set(FREETYPE_LIBRARY freetype)                 # tricking find_package into working
set(FREETYPE_INCLUDE_DIRS "${CMAKE_BINARY_DIR}/freetype/include/" "${CMAKE_CURRENT_LIST_DIR}/freetype/include/")
add_subdirectory(SDL_ttf EXCLUDE_FROM_ALL)
```
With this change, building freetype is made much simpler: 
```cmake
add_subdirectory(freetype EXCLUDE_FROM_ALL)
set(SDLTTF_FREETYPE_VENDORED ON CACHE INTERNAL "")
add_subdirectory(SDL_ttf EXCLUDE_FROM_ALL)	
```
This should also help FetchContent users.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -250,39 +250,46 @@ endif()
 
 if(SDLTTF_FREETYPE)
     if(SDLTTF_FREETYPE_VENDORED)
-        message(STATUS "${PROJECT_NAME}: Using vendored freetype library")
-        # FT_DISABLE_ZLIB variable is used by freetype
-        set(FT_DISABLE_ZLIB ON CACHE BOOL "freetype zlib option")
-        # FT_DISABLE_BZIP2 variable is used by freetype
-        set(FT_DISABLE_BZIP2 ON CACHE BOOL "freetype bzip2 option")
-        # FT_DISABLE_PNG variable is used by freetype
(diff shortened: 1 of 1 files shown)


madebr:
Aren't you using SDL_ttf with `SDLTTF_VENDORED=OFF` then?
`SDLTTF_VENDORED=ON` means "we'll provide the dependencies, you don't have to worry about this".

madebr (changes requested):
I tend to not accept this.
If you start to provide SDL_ttf's dependencies, then you need to configure SDL_ttf with `-DSDLTTF_VENDORED=OFF`.
`SDLTTF_VENDORED=ON` is provided as a easy(!) way to build SDL_ttf including all dependencies.  We either provide them all, or none at all.

PR #461, opened 2025-01-08
testgputext: Be less dependent on a bundled font

smcv:
Linux distributions generally don't like convenience copies of "stuff" that conceptually belongs to one package being included in a different package, because they cause duplication between packages, require extra copyright/license bookkeeping, and/or require maintainers to track down corresponding source code in order to meet the distro's self-imposed rules; so I'd prefer not to include a convenience copy of Noto in the Debian-packaged version of SDL_ttf. I can do most of this as Debian-specific changes, but some of them seem like they would equally make sense upstream:

* testgputext: Don't crash if the desired font cannot be found in the cwd

* testgputext: Allow a font filename to be given on the command-line
    
    This allows it to be used as an "as-installed" test, just like
    showfont and glfont, without needing to bundle a specific font file
    or be in a specific directory.

### DIFF
--- a/examples/testgputext.c
+++ b/examples/testgputext.c
@@ -328,6 +328,8 @@ int main(int argc, char *argv[])
 
     check_error_bool(TTF_Init());
     TTF_Font *font = check_error_ptr(TTF_OpenFont("NotoSansMono-Regular.ttf", 50));
+    if (!font)
+        running = false;
     TTF_SetFontWrapAlignment(font, TTF_HORIZONTAL_ALIGN_CENTER);
     TTF_TextEngine *engine = check_error_ptr(TTF_CreateGPUTextEngine(context.device));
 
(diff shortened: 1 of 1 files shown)


slouken:
Merged, thanks!

smcv:
Actually, I can't help wondering whether it would be better to drop the font and require one to be supplied as a command-line option. Using the default font is not particularly ergonomic (it requires the current working directory to be a specific place).

I notice that on #412, @captain0xff wrote, in response to some resistance to bundling fonts from @slouken:

> I actually decided to ship the font because if the font isn't monospaced the example will look quite weird and also different fonts will change the size of the text drastically due to the 3D rotation, it will be very hard to mitigate that only by changing the point size.

but I tried it with a few random proportional fonts and it looks fine to me...

slouken:
Yeah, let's drop the font. Do you want to make a PR for that?

PR #462, opened 2025-01-08
Don't bundle a font for the testgputext example

smcv:
Providing fonts is not SDL_ttf's job, and any game that wants to use it should either provide its own fonts or call into an OS library like fontconfig. For examples and manual tests, it's simplest if we require a font filename to be given as input, like showfont and glfont already did.


smcv:
As discussed on #461.

PR #466, opened 2025-01-21
Update SDL_ttf.c

MrColour:
Fix: `error: no previous prototype for function 'TTF_GetNumFontFaces' [-Werror,-Wmissing-prototypes]`

Function TTF_GetNumFontFaces was made obsolete in commit: https://github.com/libsdl-org/SDL_ttf/commit/c4383450236973c878ba84f323cc8c86751b2d36

Unsure if function is being kept around for backwards compatibility or something.


slouken:
I actually need to revert that commit. Returning the face directly isn't safe.

PR #469, opened 2025-01-24
Some minor improvements

captain0xff:
* Updated the testgputext example to use the latest best practices
* Updated the documentation to mention which TTF_SetText* functions can cause the internal text atlas to update/change.


slouken:
Merged, thanks!

PR #474, opened 2025-01-27
Find in hash table, use correct value size

1bsyl:
Hey @slouken 
This fix a crash in lastest version.  *** stack smashing detected ***: terminated
because the key have has to be a ptr size.

Besides, I tested with/without a hash-table for glyph_indice, and there is not difference in speed. So probably freetype FT_Get_Char_Index is already optimized and we need no glyph_indice table.


slouken:
Ah, thank you.

FT_Get_Char_Index() uses a binary search, which is probably fine on Latin-1 fonts but slow on asian fonts. We should probably benchmark it more later, but we can leave as-is for now.

PR #485, opened 2025-01-28
Added TTF_CreateRendererTextEngineWithProperties and TTF_CreateGPUTextEngineWithProperties

captain0xff:
The docs are missing. I will write them once someone confirms that the api is not problematic.

### DIFF
--- a/src/SDL_gpu_textengine.c
+++ b/src/SDL_gpu_textengine.c
@@ -25,8 +25,6 @@
 #include "SDL_hashtable.h"
 #include "SDL_hashtable_ttf.h"
 
-#define ATLAS_TEXTURE_SIZE 1024
-
 #define STB_RECT_PACK_IMPLEMENTATION
 #define STBRP_STATIC
 #define STBRP_SORT   SDL_qsort
@@ -75,6 +73,7 @@ typedef struct TTF_GPUTextEngineData
     SDL_GPUDevice *device;
     SDL_HashTable *fonts;
     AtlasTexture *atlas;
+    int atlas_texture_size;
     TTF_GPUTextEngineWinding winding;
 } TTF_GPUTextEngineData;
 
@@ -173,7 +172,7 @@ static void DestroyAtlas(SDL_GPUDevice *device, AtlasTexture *atlas)
     SDL_free(atlas);
 }
 
-static AtlasTexture *CreateAtlas(SDL_GPUDevice *device)
+static AtlasTexture *CreateAtlas(SDL_GPUDevice *device, int atlas_texture_size)
 {
     AtlasTexture *atlas = (AtlasTexture *)SDL_calloc(1, sizeof(*atlas));
     if (!atlas) {
@@ -184,8 +183,8 @@ static AtlasTexture *CreateAtlas(SDL_GPUDevice *device)
     info.type = SDL_GPU_TEXTURETYPE_2D;
     info.format = SDL_GPU_TEXTUREFORMAT_R8G8B8A8_UNORM;
     info.usage = SDL_GPU_TEXTUREUSAGE_SAMPLER | SDL_GPU_TEXTUREUSAGE_COLOR_TARGET;
-    info.width = ATLAS_TEXTURE_SIZE;
-    info.height = ATLAS_TEXTURE_SIZE;
+    info.width = atlas_texture_size;
+    info.height = atlas_texture_size;
     info.layer_count_or_depth = 1;
     info.num_levels = 1;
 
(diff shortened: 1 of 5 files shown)


src/SDL_renderer_textengine.c
```diff
@@ -838,10 +852,12 @@ TTF_TextEngine *TTF_CreateRendererTextEngine(SDL_Renderer *renderer)
         return NULL;
     }
 
+    int atlas_texture_size = SDL_GetNumberProperty(props, TTF_PROP_RENDERER_TEXT_ENGINE_ATLAS_TEXTURE_SIZE, 512);
```
sezero, line 855: Do we not need a sanity check here, e.g. against `atlas_texture_size > 0` ?

sezero: Adding an `(int)` cast here would be good here to silence MSVC

sezero: I assume it's intentional that you reduced default atlas size from 1024 to 512?

captain0xff: > I assume it's intentional that you reduced default atlas size from 1024 to 512?

Yes, I think now that we can set the size, a smaller default might be better.

slouken: Let's leave the default to 1024. Now that we can set the size, people can make it smaller if they want. :)

src/SDL_gpu_textengine.c
```diff
@@ -207,13 +206,13 @@ static AtlasTexture *CreateAtlas(SDL_GPUDevice *device)
...
 
-    int num_nodes = ATLAS_TEXTURE_SIZE / 4;
+    int num_nodes = atlas_texture_size / 4;
     atlas->packing_nodes = (stbrp_node *)SDL_calloc(num_nodes, sizeof(*atlas->packing_nodes));
     if (!atlas->packing_nodes) {
         DestroyAtlas(device, atlas);
         return NULL;
     }
-    stbrp_init_target(&atlas->packer, ATLAS_TEXTURE_SIZE, ATLAS_TEXTURE_SIZE, atlas->packing_nodes, num_nodes);
+    stbrp_init_target(&atlas->packer, atlas_texture_size, atlas_texture_size, atlas->packing_nodes, num_nodes);
```
slouken, line 215: Can this be a member of the atlas instead of being passed around as a parameter?

src/SDL_gpu_textengine.c
```diff
@@ -928,20 +929,40 @@ void TTF_DestroyGPUTextEngine(TTF_TextEngine *engine)
...
     }
 
     TTF_TextEngine *engine = (TTF_TextEngine *)SDL_malloc(sizeof(*engine));
     if (!engine) {
+        SDL_SetError("Failed to create GPU text engine.");
+        return NULL;
+    }
+
+    int atlas_texture_size = (int)SDL_GetNumberProperty(props, TTF_PROP_GPU_TEXT_ENGINE_ATLAS_TEXTURE_SIZE, 512);
+    if (atlas_texture_size <= 0) {
```
slouken, line 957: Should zero mean the default?

src/SDL_hashtable_ttf.c
```diff
@@ -54,7 +54,7 @@ static void SDL_NukeFreeGlyphHashtableKey(const void *key, const void *value, vo
 
 SDL_HashTable *SDL_CreateGlyphHashTable(SDL_GlyphHashTable_NukeFn nukefn)
 {
-    const int num_buckets = 4; // FIXME: Is this a good value?
+    const int num_buckets = 32; // FIXME: Is this a good value?
```
slouken, line 57: I'll make this change separately.

src/SDL_renderer_textengine.c
```diff
@@ -828,20 +829,40 @@ static void SDLCALL DestroyText(void *userdata, TTF_Text *text)
...
+    SDL_Renderer *renderer = SDL_GetPointerProperty(props, TTF_PROP_RENDERER_TEXT_ENGINE_RENDERER, NULL);
     if (!renderer) {
-        SDL_InvalidParamError("renderer");
+        SDL_SetError("Failed to create renderer text engine: Invalid renderer.");
         return NULL;
     }
 
     TTF_TextEngine *engine = (TTF_TextEngine *)SDL_malloc(sizeof(*engine));
     if (!engine) {
+        SDL_SetError("Failed to create renderer text engine.");
```
slouken, line 852: You should leave the "Out of memory" error that's set by SDL_malloc()

src/SDL_gpu_textengine.c
captain0xff: Is there any reason people will want to set the property to 0 particularly? Because otherwise I think handling 0 as an error might help to track potential bugs in code.

src/SDL_renderer_textengine.c
captain0xff: Reverted

slouken (approved):
This looks good. Can you resolve the conflicts and I'll merge it?

captain0xff:
Did this sync fix it? Because it's showing me that the branch has no conflicts.

slouken:
Yep, thanks!

PR #486, opened 2025-01-28
Fixed SDF text layout

slouken:
We're also passing through the raw SDF values now so shaders can be used to render them correctly.


sezero:
Are these changes needed in SDL2 (and thus release-2.24.x) branches?

slouken:
No, these are not backwards compatible changes.

PR #506, opened 2025-02-06
Implemented setting font features and font variations. (#503)

### LINKED ISSUE #503: [Feature request] Setting font features and font variations.
The ability to enable or disable font features would be a useful addition. I'm making a text editor, and I want to support the user changing font features and variations to their preference.

Font features are things like ligatures, '[texture healing](https://github.com/githubnext/monaspace?tab=readme-ov-file#texture-healing)', and alternate characters (slashed vs dotted zero). Font variations are things like font width, font **weight** and font _slant_.

As an example, the MonaSpace family of fonts (https://monaspace.githubnext.com/), supports many font features and varations.

Here is the showfont demo, using the Monaspace Argon font. The left is with all the stylistic sets (ligatures) , the middle is with the default font features, and the right is with "-calt" which disables font healing (see how the `m`'s are no longer different widths depending on what characters come before and after it):

Image

### API ideas
```c
TTF_SetFontFeatures(font, "calt liga=1 -kern ss01 aalt[3:5]=2"); // space-separated features
TTF_SetFontVariations(font, "wght=500 slnt=-7.5"); // space-separated variations
```

(the format for each feature/variation are just what harfbuzz supports, see `hb_feature_from_string` and `hb_variation_from_string` in https://harfbuzz.github.io/harfbuzz-hb-common.html )

Should there be `TTF_GetFontFeatures` and `TTF_GetFontVariations` as well?

How should enabling or disabling kerning using `TTF_SetFontFeatures` interact with `TTF_SetFontKerning`?

There are essentially an infinite (I think 36^4, a-z0-9) number of font features a user can specify. For example, there are 100 character variants, `cv00`-`cv99` a font can chose to add so implementing this will require dynamic memory allocation, unless we set an arbitrary limit on how many features can be specified?

For font variations, there are only 5 ([registered opentype variations](https://learn.microsoft.com/en-us/typography/opentype/spec/dvaraxisreg#registered-axis-tags)), at least currently. I don't know enough about fonts to say if there are font formats other than OpenType which support these features. It's not documented that I can see in HarfBuzz. MDN has an example of using "GRAD" as a custom font variation. I don't think it changes rendering though, but if custom font variations are a thing, we probably shouldn't limit this one to the five registered variations?

### Implementation

If this is something you want added to the library, I can attempt to implement this. I've implemented `TTF_SetFontFeatures`, which seems to work with basic testing. I can open a draft PR once I've implemented the other functions, if you want me to.

Martin-Eriksson:
Closes #503.

I've got some comments in there with questions about how to do certain things.

I had thought I could use `hb_font_set_variations` and it would be super easy, 
barely an inconvenience, to support setting variations but it does nothing at all.
Instead I had to use FreeType directly, which took me a while to figure out how
to do. Do you think `hb_font_set_variations` is supposed to work? Or is it correct
that it does nothing when using FreeType, perhaps?

`font->enable_kerning` is not used at all anymore, `TTF_SetFontFeatures` and `TTF_SetFontKerning`
are completely independent. It ignores the `font->enable_kerning` value 
if `TTF_SetFontFeatures` has been called with a non-null string. If `TTF_SetFontFeatures` 
is called later with a null string, it starts using the `font->enable_kerning` value again.
Combining them is a bit complicated if we want to avoid doing more work in `CollectGlyphsFromFont()`.

I've tried to run address sanitizer on the code but I haven't got it working yet.
It reports nothing even though I've commented out calls to free.

Until I've figured out how to get ASan working, and if `hb_font_set_variations` is supposed to work, I'm keeping this as 
a draft PR

### DIFF
--- a/src/SDL_ttf.c
+++ b/src/SDL_ttf.c
@@ -32,6 +32,8 @@
 #include FT_GLYPH_H
 #include FT_TRUETYPE_IDS_H
 #include FT_IMAGE_H
+#include FT_MULTIPLE_MASTERS_H
+#include FT_SFNT_NAMES_H 
 
 /* Enable rendering with color
  * Freetype may need to be compiled with FT_CONFIG_OPTION_USE_PNG */
@@ -271,6 +273,14 @@ typedef struct CachedGlyphPositions {
     GlyphPositions positions;
 } CachedGlyphPositions;
 
+//naming? it's not from FreeType so maybe FT_ is inappropriate?
+typedef struct FT_Variations { 
+    int len;
+    FT_ULong *tags;
+    FT_Fixed *defaults;
+    FT_Fixed *coords;
+} FT_Variations;
+
 // A structure maintaining a list of fonts
 typedef struct TTF_FontList {
     TTF_Font *font;
@@ -345,6 +355,15 @@ struct TTF_Font {
 #if TTF_USE_HARFBUZZ
     hb_font_t *hb_font;
     hb_language_t hb_language;
+
+    char *features;
+    hb_feature_t *hb_features;
+    int hb_features_len;
+
+    char *variations;
+    hb_variation_t *hb_variations;
+    int hb_variations_len;
+    FT_Variations ft_variations;
 #endif
     Uint32 script;
     TTF_Direction direction;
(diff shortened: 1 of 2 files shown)


slouken:
@Martin-Eriksson, I'd like to wrap up an SDL_ttf release, what's the status of this PR? We can definitely add these functions later, but if you'd like to get them in, I'm fine with that.

RobertSundling:
I noticed some compiler warnings when testing this PR with GCC. I've fixed these warnings in my fork at:
https://github.com/RobertSundling/SDL_ttf/tree/fix-pr506-compiler-warnings

Feel free to incorporate these fixes into this PR if you find them useful.

src/SDL_ttf.c
```diff
@@ -3358,13 +3377,35 @@ static bool CollectGlyphsFromFont(TTF_Font *font, const char *text, size_t lengt
...
-    userfeatures[0].value = font->enable_kerning;
-    userfeatures[0].start = HB_FEATURE_GLOBAL_START;
-    userfeatures[0].end = HB_FEATURE_GLOBAL_END;
+    if (font->hb_variations) {
+#if 0
+        //! hb_font_set_variations does nothing
+        //! hb_font_set_variation doesn't work either
+        // is it supposed to work? Is it a bug in HarfBuzz or does it not work
+        // because we're using FreeType?
+        hb_font_set_variations(font->hb_font, font->variations, font->variations_len);
```
mcendu: Try calling [`hb_ft_hb_font_changed()`](https://harfbuzz.github.io/harfbuzz-hb-ft.html#hb-ft-hb-font-changed) after this?

While you no longer need to call it in HarfBuzz 11.0.0 or later, not all distributions ship this new a HarfBuzz.

src/SDL_ttf.c
```diff
@@ -6086,6 +6127,211 @@ bool TTF_SetFontLanguage(TTF_Font *font, const char *language_bcp47)
...
+}
+
+char *TTF_GetFontFeatures(TTF_Font *font)
+{
+    TTF_CHECK_FONT(font, false);
+
+#if TTF_USE_HARFBUZZ
+    return font->features;
+#else
+    (void) features;
```
mcendu: This line is unnecessary.

src/SDL_ttf.c
```diff
@@ -3358,13 +3377,35 @@ static bool CollectGlyphsFromFont(TTF_Font *font, const char *text, size_t lengt
...
+#if 0
+        //! hb_font_set_variations does nothing
+        //! hb_font_set_variation doesn't work either
+        // is it supposed to work? Is it a bug in HarfBuzz or does it not work
+        // because we're using FreeType?
+        hb_font_set_variations(font->hb_font, font->variations, font->variations_len);
+#else
+        // should we care if this returns an error? what should we do in that case?
+        FT_Set_Var_Design_Coordinates(font->face, font->ft_variations.len, font->ft_variations.coords);
+#endif
```
mcendu: Try calling [hb_ft_hb_font_changed()](https://harfbuzz.github.io/harfbuzz-hb-ft.html#hb-ft-hb-font-changed) after [hb_font_set_variations()](https://harfbuzz.github.io/harfbuzz-hb-font.html#hb-font-set-variations)?

While you no longer need to call it in HarfBuzz 11.0.0 or later, not all distributions ship this new a HarfBuzz.


```suggestion
        hb_font_set_variations(font->hb_font, font->variations, font->variations_len);
        hb_ft_hb_font_changed(font->hb_font);
```

mcendu (changes requested):
Are you still working on this? If not, can I take this over?

Also what were you doing when ASan threw the error, and what did ASan print?

PR #525, opened 2025-03-09
Fix missing option that causes PlutoSVG examples to be added to the build system

Ravbug:
`svg2png` and `camera2png` were being added to the build system


madebr (approved):

slouken:
Merged, thanks!

PR #528, opened 2025-03-17
ASAN Issue

tanolino:
On MSVC I get a
```
==9952==ERROR: AddressSanitizer: memcpy-param-overlap: memory ranges [0x11e8718f6e00,0x11e8718f6e78) and [0x11e8718f6e18, 0x11e8718f6e90) overlap
```
Since the CRT `memcpy` memory is not allowed to overlap (see: https://learn.microsoft.com/en-us/cpp/sanitizers/error-memcpy-param-overlap?view=msvc-170 ) And `SDL_memcpy` uses `memcpy` by default.

This fix might be sufficient.
Another alternative is to use `memmove` to move overlapping memory.

### DIFF
--- a/src/SDL_renderer_textengine.c
+++ b/src/SDL_renderer_textengine.c
@@ -367,7 +367,9 @@ static bool ResolveMissingGlyphs(TTF_RendererTextEngineData *enginedata, AtlasTe
             // Remove this from the missing entries
             --num_missing;
             if (i < num_missing) {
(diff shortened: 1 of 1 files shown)


slouken:
Yeah, let’s switch this to SDL_memmove(), thanks!

tanolino:
Took me a few minutes to find out how to do that in here.
I have swapped it.

slouken:
Merged, thanks!

Issue #527, opened 2025-03-17
[Feature request] Retrieving the weight of a font.


TriangulumDesire:
I am suggesting to add some functions that query the weight (in other words, the lightness/heaviness) of a font. This can be used to help the user disambiguate different fonts within the same family.

For example, in one of my applications, I have a `FontFamily` class that has a collection of fonts that belong to the same family. The user can then supply a weight and the `FontFamily` class will return the most relevant font face.

A further suggestion (that depends on this one) is named font weights, such as the ones used in [WPF](https://learn.microsoft.com/en-us/dotnet/api/system.windows.fontweights?view=windowsdesktop-9.0) and [CSS](https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight#common_weight_name_mapping).

To summarise, I propose the following API:
```c
typedef enum TTF_FontWeight
{
    TTF_FONT_WEIGHT_UNKNOWN = 0,
    TTF_FONT_WEIGHT_THIN = 100,
    TTF_FONT_WEIGHT_EXTRA_LIGHT = 200,
    TTF_FONT_WEIGHT_LIGHT = 300,
    TTF_FONT_WEIGHT_NORMAL = 400,
    TTF_FONT_WEIGHT_MEDIUM = 500,
    TTF_FONT_WEIGHT_SEMI_BOLD = 600,
    TTF_FONT_WEIGHT_BOLD = 700,
    TTF_FONT_WEIGHT_EXTRA_BOLD = 800,
    TTF_FONT_WEIGHT_BLACK = 900,
    TTF_FONT_WEIGHT_EXTRA_BLACK = 950,
} TTF_FontWeight;

int TTF_GetFontWeight(const TTF_Font *font);
TTF_FontWeight TTF_GetFontWeightNamed(const TTF_Font *font);
```

Let me know if this is something you'd like to see in this library, and if there are any changes you'd made to my propose API.

I am more than happy to make a pull request for this once you're satisfied with the proposal.

slouken:
This sounds interesting, what would the implementation look like?

TriangulumDesire:
The font's weight is stored in the `usWeightClass` of the [OS/2 table](https://learn.microsoft.com/en-us/typography/opentype/spec/os2#usweightclass) and its values are usually between 1 and 1,000. Retrieval in FreeType would look something like this:

```c
int font_weight = TTF_FONT_WEIGHT_UNKNOWN;
[...]

PR #531, opened 2025-03-18
Add font weight functionality

TriangulumDesire:
Solves #527.

Adds `weight` to `TTF_Font` (which is retrieved from a font's [OS/2 table](https://learn.microsoft.com/en-us/typography/opentype/spec/os2)), as well as `TTF_GetFontWeight`—which is used to retrieve this value.

Also adds a set of `#define`s that correspond to [commonly-used named font weight values](https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight#common_weight_name_mapping). These can be used for ‘categorising’ fonts to certain weight classes, and will also be useful for setting the `wght` value for font variations (#506).

This pull request also fixes a minor documentation comment error.

### DIFF
--- a/src/SDL_ttf.c
+++ b/src/SDL_ttf.c
@@ -31,6 +31,7 @@
 #include FT_STROKER_H
 #include FT_GLYPH_H
 #include FT_TRUETYPE_IDS_H
+#include FT_TRUETYPE_TABLES_H
 #include FT_IMAGE_H
 
 /* Enable rendering with color
@@ -306,6 +307,7 @@ struct TTF_Font {
 
     // The font style
     TTF_FontStyleFlags style;
+    int weight;
     int outline;
     FT_Stroker stroker;
 
@@ -2129,6 +2131,7 @@ TTF_Font *TTF_OpenFontWithProperties(SDL_PropertiesID props)
     // Set the default font style
     if (existing_font) {
         font->style = existing_font->style;
+        font->weight = existing_font->weight;
         font->outline = existing_font->outline;
         font->ft_load_target = existing_font->ft_load_target;
         font->enable_kerning = existing_font->enable_kerning;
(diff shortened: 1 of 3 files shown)


src/SDL_ttf.c
```diff
@@ -2137,6 +2140,16 @@ TTF_Font *TTF_OpenFontWithProperties(SDL_PropertiesID props)
         font->outline = 0;
         font->ft_load_target = FT_LOAD_TARGET_NORMAL;
         TTF_SetFontKerning(font, true);
+
+        // Retrieve the weight from the OS2 TrueType table
+        const TT_OS2 *os2_table = (const TT_OS2 *)FT_Get_Sfnt_Table(face, FT_SFNT_OS2);
+        if (os2_table != NULL && os2_table->usWeightClass != 0) {
+            font->weight = os2_table->usWeightClass;
+        } else if (face->style_flags & FT_STYLE_FLAG_BOLD) {
```
slouken, line 2148: The font weight should probably be recalculated when font styles change.

TriangulumDesire: I don't think that the weight will be affected by any style changes. Unless I'm missing something, the dependent values `os2_table->usWeightClass` and `face->style_flags` will remain constant, regardless of any changes to the font's style.

However, looking into this, I realised that SDL_ttf's custom `TTF_STYLE_BOLD` rendering will output glyphs whose weight is different than that of the font's. If we did want to reflect this boldness in the weight value, then I'd definitely have to (re)calculate an adjusted weight value when the font style changes for any fonts with the `TTF_STYLE_BOLD` flag. However, I don't think it is typographically quantifiable, since the custom boldness appears to render the glyphs twice with a horizontal offset, instead of changing the ‘weight’ of the rendered glyphs.

PR #538, opened 2025-03-28
Add TTF_GetFontBBox() which returns font's extremum bounding box

### LINKED ISSUE #536: [Feature request]: query font's BBOX
I've met few situations where a font has incorrect ascent/descent information. I'm not an expert in this, and don't know how this usually happens, either it's saved incorrectly, or font's author defined them so for a reason. In any case, this makes the ascent/descent, and as a consequence - height values to not give an actual representation of how big the font letters are.

Here's one font example that I have, if necessary:
[Pixel-Noir Skinny Short.zip](https://github.com/user-attachments/files/19399787/Pixel-Noir.Skinny.Short.zip)
If imported at point size 14 this font reports ascent 4, descent 1, and therefore height as 3 (sic), but the actual graphic height of the glyphs would be ~26 pixels.

I learnt that FT_Face has a field called "bbox", which is *"The box is large enough to contain any glyph from the font."*. This lets to find the real graphical height of this font using following code:

```
FT_Long real_ascent = FT_MulFix(FT_DivFix(ft_face->bbox.yMax, ft_face->units_per_EM), ft_face->size->metrics.y_ppem);
FT_Long real_descent = FT_MulFix(FT_DivFix(ft_face->bbox.yMin, ft_face->units_per_EM), ft_face->size->metrics.y_ppem);
```

Would it be possible to add a function that returns font's bbox, or at least a vertical component of bbox (yMax, yMin), preferably already converted to scaled units?

ivan-mogilko:
Resolve #536

This adds a new function called TTF_GetFontBBox:
```
extern DECLSPEC void SDLCALL TTF_GetFontBBox(const TTF_Font *font, TTF_BBox *bbox);
```
and a struct TTF_BBox (simply a pair of xmin/max and ymin/max offsets).

This function may be used to query font's bbox, which is a bounding box enough to suit any glyph from this font.
Function description is based on FreeType comments to FT_Face's bbox field.

I was not certain which version to mention in comments, so put 2.26.0, as that seems to be the next release number, theoretically?

### DIFF
--- a/SDL_ttf.c
+++ b/SDL_ttf.c
@@ -3040,6 +3040,27 @@ void TTF_SetFontKerning(TTF_Font *font, int allowed)
 #endif
 }
 
+void TTF_GetFontBBox(const TTF_Font *font, int *minx, int *maxx, int *miny, int *maxy)
+{
+    if (!font || !font->face) {
+        return;
+    }
+    /* Recalculate FT_Face's bbox from font units to pixels */
+    const FT_Face face = font->face;
+    if (minx) {
+        *minx = FT_MulFix(FT_DivFix(face->bbox.xMin, face->units_per_EM), face->size->metrics.x_ppem);
+    }
+    if (maxx) {
+        *maxx = FT_MulFix(FT_DivFix(face->bbox.xMax, face->units_per_EM), face->size->metrics.x_ppem);
+    }
+    if (miny) {
+        *miny = FT_MulFix(FT_DivFix(face->bbox.yMin, face->units_per_EM), face->size->metrics.y_ppem);
+    }
+    if (maxy) {
+        *maxy = FT_MulFix(FT_DivFix(face->bbox.yMax, face->units_per_EM), face->size->metrics.y_ppem);
+    }
+}
+
 long TTF_FontFaces(const TTF_Font *font)
 {
     return font->face->num_faces;
--- a/SDL_ttf.h
+++ b/SDL_ttf.h
@@ -679,6 +679,22 @@ extern DECLSPEC int SDLCALL TTF_GetFontKerning(const TTF_Font *font);
 extern DECLSPEC void SDLCALL TTF_SetFontKerning(TTF_Font *font, int allowed);
 
 /**
+ * Query the font bounding box.
+ * 
+ * The bounding box defines bounds large enough to contain any glyph from
+ * the font. It is expressed in pixel offsets from glyph's origin (0,0),
+ * with Y axis pointing upwards. Thus ymax offset may be seen as the
+ * "maximum ascender" and ymin offset - as the "minimum descender".
+ *
+ * \param font the font to query.
+ *
+ * \since This function is available since SDL_ttf 2.26.0.
+ */
+extern DECLSPEC void SDLCALL TTF_GetFontBBox(const TTF_Font *font,
+                        int *minx, int *maxx,
+                        int *miny, int *maxy);
+
+/**
  * Query the number of faces of a font.
  *
  * \param font the font to query.


SDL_ttf.h
```diff
@@ -678,6 +687,21 @@ extern DECLSPEC int SDLCALL TTF_GetFontKerning(const TTF_Font *font);
...
+ * the font. It is expressed in pixel offsets from glyph's origin (0,0),
+ * with Y axis pointing upwards. Thus ymax offset may be seen as the
+ * "maximum ascender" and ymin offset - as the "minimum descender".
+ *
+ * \param font the font to query.
+ * \param bbox the TTF_BBox struct to fill.
+ *
+ * \since This function is available since SDL_ttf 2.26.0.
+ */
+extern DECLSPEC void SDLCALL TTF_GetFontBBox(const TTF_Font *font, TTF_BBox *bbox);
```
slouken, line 703: Rather than creating a new type, can you have separate pointers for each element that can be optionally filled in?

SDL_ttf.h
```diff
@@ -678,6 +687,21 @@ extern DECLSPEC int SDLCALL TTF_GetFontKerning(const TTF_Font *font);
  */
 extern DECLSPEC void SDLCALL TTF_SetFontKerning(TTF_Font *font, int allowed);
 
+/**
+ * Query the font bounding box.
+ * 
+ * The bounding box defines bounds large enough to contain any glyph from
+ * the font. It is expressed in pixel offsets from glyph's origin (0,0),
+ * with Y axis pointing upwards. Thus ymax offset may be seen as the
```
slouken, line 695: This is the opposite of how coordinates are usually done in SDL. Can you double check the other SDL_ttf APIs and see if that's consistent here?

SDL_ttf.c
```diff
@@ -3035,6 +3035,16 @@ void TTF_SetFontKerning(TTF_Font *font, int allowed)
 #endif
 }
 
+void TTF_GetFontBBox(const TTF_Font *font, TTF_BBox *bbox)
+{
+    /* Recalculate FT_Face's bbox from font units to pixels */
```
slouken, line 3040: You should make sure the font is valid, see the other API functions for examples.

SDL_ttf.h
ivan-mogilko: By "opposite", are you referring to Y axis?

SDL_ttf.h
ivan-mogilko: Changed the prototype following `TTF_GlyphMetrics` as an example.

SDL_ttf.c
ivan-mogilko: I guess I have to test that `font` pointer is valid? By the way, I noticed that not all functions have this check, only some do.

SDL_ttf.h
ivan-mogilko: I can confirm that TTF_FontAscent and TTF_FontDescent functions return these offsets in a similar style: ascent is normally a positive value and descent is a negative value, as if Y axis is pointing up. For properly done fonts their return values also match miny and maxy values returned from TTF_GetFontBBox. (Their values differ for the fonts which have incorrect metrics information in them.)

ivan-mogilko: By the way, since there's a TTF_GlyphMetrics function, do you think that TTF_GetFontBBox is a okay name, or should I rename to something more consistent with "GlyphMetrics"? Because the meaning of BBox is essentially "max known glyph metrics".

slouken: Okay, thanks for checking, that seems fine to me. I'm not sure on the name, let me think about it a bit. What's the use case for this function?

ivan-mogilko: > What's the use case for this function?

In my case (mentioned in #536), there are fonts that have incorrect ascent/descent information for some reason (maybe because font's creator made a mistake), but there's still a way to tell how big their characters are using this bounding box info. This helps if we want to precalculate a necessary texture size before first render happened, for example.

ivan-mogilko:
@slouken are there doubts about the function's name, or the function's purpose?
I could rename it to `TTF_MaxGlyphMetrics`, for example.

slouken:
We're not currently adding new functions to the SDL2 API. Please feel free to retarget this to main and provide a concrete use case for it, as well as clarification on what measurement functions should be called in what situations if you'd like it to be merged. I'll go ahead and close it for now.

PR #559, opened 2025-06-11
Remove a misaligned pointer cast in BG_Blended_LCD

marius-jaegler:
The BG_Blended_LCD function uses Uint32 pointers to keep track of source and destination but this can cause undefined behavior if src becomes misaligned. This can happen is `srcskip` is not a multiple of `sizeof(Uint32)`. This is why the function uses `SDL_memcpy` for reading the `src` buffer.

But it is actually undefined behavior to even perform the cast on line 552 when the result is misaligned. The error is caught at runtime with ubsan in debug mode (I found it compiling with `zig` so it can also probably be found with `clang`).

From the C11 standard (C11 (n1570) 6.3.2.3 p7)
```
A pointer to an object type may be converted to a pointer to a different object type. If the
resulting pointer is not correctly aligned68) for the referenced type, the behavior is
undefined.
```

This patch makes the pointer a byte pointers and makes the increment sizes explicit.

There may be a similar problem with `dst` / `dskskip`  but it looks like they are always aligned in the current implementation, (as long as the image pitch is). Maybe an assertion could be added for future proofing.

### DIFF
--- a/src/SDL_ttf.c
+++ b/src/SDL_ttf.c
@@ -456,7 +456,7 @@ static bool Find_GlyphByIndex(TTF_Font *font, FT_UInt idx, int want_bitmap, int 
 // Blend colored glyphs
 static void BG_Blended_Color(const TTF_Image *image, Uint32 *destination, Sint32 srcskip, Uint32 dstskip, Uint8 fg_alpha)
 {
-    const Uint32 *src   = (Uint32 *)image->buffer;
+    const Uint8  *src   = image->buffer;
     Uint32      *dst    = destination;
     Uint32       width  = image->width;
     Uint32       height = image->rows;
@@ -465,10 +465,12 @@ static void BG_Blended_Color(const TTF_Image *image, Uint32 *destination, Sint32
         while (height--) {
             /* *INDENT-OFF* */
             DUFFS_LOOP4(
-                *dst++ = *src++;
+                SDL_memcpy(dst, src, sizeof(Uint32));
+                src += sizeof(Uint32);
+                dst += sizeof(Uint32);
             , width);
             /* *INDENT-ON* */
-            src = (const Uint32 *)((const Uint8 *)src + srcskip);
+            src += srcskip;
             dst = (Uint32 *)((Uint8 *)dst + dstskip);
         }
     } else {
@@ -480,7 +482,8 @@ static void BG_Blended_Color(const TTF_Image *image, Uint32 *destination, Sint32
             DUFFS_LOOP4(
                     /* prevent misaligned load: tmp = *src++; */
                     // eventually, we can expect the compiler to replace the memcpy call with something optimized
-                    SDL_memcpy(&tmp, src++, sizeof(tmp));
+                    SDL_memcpy(&tmp, src, sizeof(tmp));
+                    src += sizeof(Uint32);
                     alpha = tmp >> 24;
                     tmp &= ~0xFF000000;
                     alpha = fg_alpha * alpha;
@@ -488,7 +491,7 @@ static void BG_Blended_Color(const TTF_Image *image, Uint32 *destination, Sint32
                     *dst++ = tmp | alpha
                     , width);
             /* *INDENT-ON* */
-            src = (const Uint32 *)((const Uint8 *)src + srcskip);
+            src += srcskip;
             dst = (Uint32 *)((Uint8 *)dst + dstskip);
         }
     }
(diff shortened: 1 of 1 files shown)


slouken:
It might be worth seeing in what cases you get an unaligned pointer. I know the start of the images are always aligned by definition. It might be better for the SIMD functions to just fall back to non-SIMD functions in the unaligned case.

marius-jaegler:
Actually it looks like the SIMD versions are only ever used with srcskip 32bit aligned (as long as the image pitches are multiple of 4), so those might no need to really be changed

marius-jaegler:
The other option is to just not use the SIMD versions if srcskip is misaligned, which should be pretty easy

1bsyl:
Hi, 

There was so old ticket about this:
https://github.com/libsdl-org/SDL_ttf/issues/324
https://github.com/libsdl-org/SDL_ttf/pull/329
They were fix for sparc cpu's.

IIRC, neon allows unaligned load/store, it will only be slower.

otherwise, does it crash somewhere, or just an ubsan warning ?

marius-jaegler:
Wether it is a crash or just a warning depends on how ubsan is configured (it was a crash in my case), but yes, it is probably something that only sanitizers catch.

If neon allows unaligned load/store at the byte level like the SSE functions, it should be possible to apply the exact same fix to the NEON version where we wrap the incorrectly typed intrinsic into an inline function with sanitization disabled (which would really only be necessary if there exists a sanitizer that can catch it for ARM platforms)

1bsyl:
> Wether it is a crash or just a warning depends on how ubsan is configured (it was a crash in my case), but yes, it is probably something that only sanitizers catch.
> 

I mean't without ubsan. So no crash in real situation.

Though, any misaligned access should crash on sparc. So all BG_ function for lcd/blended/color, etc. without simd, neon.

> If neon allows unaligned load/store at the byte level like the SSE functions, it should be possible to apply the exact same fix to the NEON version where we wrap the incorrectly typed intrinsic into an inline function with sanitization disabled (which would really only be necessary if there exists a sanitizer that can catch it for ARM platforms)

I dont get it. if new allows unaligned load/store, what do we have to fix anything ?

marius-jaegler:
The reason the fix is needed is that the problem is that it is not just a misaligned load/store that is undefined behavior but also the cast itself, for example with the non-SIMD version, an optimizing compiler would be allowed to vectorize the loop itself using any SIMD instructions available and it would be allowed to assume that the pointer is aligned correctly for it's type (which is not true in this case).

Or for something simpler, with the current load where `tmp = *src` is replaced with `SDL_memcpy(&tmp, src, 4)`, since SDL_memcpy is actually just a define for memcpy, the compiler is allowed to see that because `src` is of type `Uint32*`, it must be aligned to 4 bytes, which means that the memcpy call can be replaced by a simple move and that would  cause a crash on an architecture where misaligned loads are illegal. Although I don't know if there is currently any compiler that actually does this optimization.

For the SSE/NEON, the "fix" mostly just makes ubsan shut up while still being technically undefined behavior as far as the C standard is concerned.

1bsyl:
This wasn't much tested it seems.
because it crashed rendering any emoji / with color blended.(see valgrind log).
fixed in https://github.com/libsdl-org/SDL_ttf/commit/a9a4fea81b41f3ceefcfacd5b32fa3e838a554c0

there is a test program with a fuzzer/random string rendering, which should never crash.
examples/testapp.c
you need to add you fonts to test. F1 for help. press 'r' to run the random rendering



```
TTF_OpenFont: fonts/NotoColorEmoji.ttf
 29701] seed=1752609666; replay=1; font_style=1; kerning=0; sdf=0; wrap=0; wrap_size=61; w_align=2; outline=0; curr_size=6; render_mode=1; curr_str=57; curr_font=0; hinting=1; fg_alpha=0; // light Blended
==38175== Invalid write of size 4
==38175==    at 0x487CCA3: memcpy (string_fortified.h:29)
==38175==    by 0x487CCA3: BG_Blended_Color.isra.0 (SDL_ttf.c:467)
==38175==    by 0x4882F52: Render_Line_SSE_Blended_Opaque (SDL_ttf.c:1323)
==38175==    by 0x4882F52: Render_Line (SDL_ttf.c:1427)
==38175==    by 0x4883887: TTF_Render_Internal (SDL_ttf.c:3949)
==38175==    by 0x10BD07: main (testapp.c:971)
==38175==  Address 0x1e3c286c is 4 bytes after a block of size 2,394,152 alloc'd
==38175==    at 0x484A858: malloc (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==38175==    by 0x4B6E046: SDL_malloc_REAL (SDL_malloc.c:6452)
==38175==    by 0x4B6FFBF: SDL_aligned_alloc_REAL (SDL_stdlib.c:546)
==38175==    by 0x487FB01: AllocateAlignedPixels (SDL_ttf.c:1636)
==38175==    by 0x4883A1F: Create_Surface_Blended (SDL_ttf.c:1762)
==38175==    by 0x4883A1F: TTF_Render_Internal (SDL_ttf.c:3939)
==38175==    by 0x10BD07: main (testapp.c:971)
```

PR #560, opened 2025-06-29
Return the errors from GetWrappedLines in LayoutText instead of ignoring them

SimonMaracine:
`GetWrappedLines` can return with a bunch of errors and `LayoutText` was previously ignoring them by returning `true`.

I stumbled into this when I made a mistake misusing the API, that caused TrueType to return an error somewhere deep into the call stack of SDL_ttf. Before the patch, that error was not showing up, instead the text was silently not being rendered.


slouken:
Merged, thanks!

PR #577, opened 2025-08-11
Implement TTF_SetFontCharSpacing()

### LINKED ISSUE #561: [Question] A custom Character Spacing
Hello, this is not exactly a feature request, but rather a sort of inquiry. I'd like to know if it's *hypothetically* possible for SDL_ttf to support custom character spacing, that works as a extra value applied to each glyph's "advance"? The big text editors, such as "MS Word" or "Writer" from Libre Office do support such behavior, so it's seems standard enough (for the text editors at least), but on another hand, I realize that this may be complicated to support overall as different languages may require different approach to how glyphs are bound together (considering e.g. ligatures, text direction, etc).

Is this something that hypothetically may be wanted or appear in SDL_ttf?
Do you think that it may be acceptable to have a "minimal" implementation of such feature where it would e.g. work consistently only for the simple fonts / languages? (Unfortunately, I am not well aware of how rendering is done within SDL_ttf, so cannot yet myself tell if it's possible to switch between such behavior dynamically.)

ivan-mogilko:
Resolves #561

This is an experimental proposal, which adds an optional uniform character spacing, applied after each drawn glyph *in addition* to the normal glyph's advance. Spacing may be both positive or negative. Setting this to 0 ensures that the font is drawn using only its kerning.

A function declared as:
```
int TTF_SetFontCharSpacing(TTF_Font *font, int char_spacing);
```


slouken:
Please retarget the main branch for this PR. It will be back-ported if necessary.

ivan-mogilko:
I shall remake this for main then, and make a new PR. The code that measures and draws text is different there, I'll need more time.

Also, I found there's a "showfont" program along with the library, might use that to test this feature too.

PR #578, opened 2025-08-12
Implement TTF_SetFontCharSpacing()

### LINKED ISSUE #561: [Question] A custom Character Spacing

ivan-mogilko:
Resolves #561

This is an experimental proposal, which adds an optional uniform character spacing, applied after each drawn glyph *in addition* to the normal glyph's advance. Spacing may be both positive or negative. Setting this to 0 ensures that the font is drawn using only its kerning.

Two new functions declared as:
```
bool TTF_SetFontCharSpacing(TTF_Font *font, int char_spacing);
int TTF_GetFontCharSpacing(TTF_Font *font);
```

Character spacing value is in pixels on input, but stored as a 26.6 FP in the TTF_Font struct, for compatibility with GlyphPosition's fields. When calculating glyph positions, "char_spacing" is added to each pos's x_advance.

Additionally, added two new key controls to the "showfont" example:
Ctrl + , - decreases character spacing
Ctrl + . - increases character spacing

### DIFF
--- a/include/SDL3_ttf/SDL_ttf.h
+++ b/include/SDL3_ttf/SDL_ttf.h
@@ -1003,7 +1003,7 @@ extern SDL_DECLSPEC bool SDLCALL TTF_SetFontCharSpacing(TTF_Font *font, int spac
  * \threadsafety This function should be called on the thread that created the
  *               font.
  *
- * \since This function is available since SDL_ttf 3.0.0.
+ * \since This function is available since SDL_ttf 3.4.0.
  */
 extern SDL_DECLSPEC int SDLCALL TTF_GetFontCharSpacing(TTF_Font *font);
 
--- a/include/SDL3_ttf/SDL_ttf.h
+++ b/include/SDL3_ttf/SDL_ttf.h
@@ -987,7 +987,7 @@ extern SDL_DECLSPEC TTF_Direction SDLCALL TTF_GetFontDirection(TTF_Font *font);
  * \threadsafety This function should be called on the thread that created the
  *               font.
  *
- * \since This function is available since SDL_ttf 3.0.0.
+ * \since This function is available since SDL_ttf 3.4.0.
  */
 extern SDL_DECLSPEC bool SDLCALL TTF_SetFontCharSpacing(TTF_Font *font, int spacing);
 
--- a/include/SDL3_ttf/SDL_ttf.h
+++ b/include/SDL3_ttf/SDL_ttf.h
@@ -970,6 +970,44 @@ extern SDL_DECLSPEC bool SDLCALL TTF_SetFontDirection(TTF_Font *font, TTF_Direct
 extern SDL_DECLSPEC TTF_Direction SDLCALL TTF_GetFontDirection(TTF_Font *font);
 
 /**
+ * Set additional space in pixels to be applied between any two rendered
+ * characters. The spacing value is applied uniformly after each character,
+ * in addition to the normal glyph's advance.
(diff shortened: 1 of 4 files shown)


1bsyl:
I think 1 glyph isn't one character in all languages.
This works for latin, but not for more complicated language arabic, etc.
also ligature chars ?

HarfBuzz should do the letter-spacing ..  
maybe that's the current state: https://github.com/typst/typst/pull/6161

include/SDL3_ttf/SDL_ttf.h
```diff
@@ -969,6 +969,44 @@ extern SDL_DECLSPEC bool SDLCALL TTF_SetFontDirection(TTF_Font *font, TTF_Direct
...
+ *
+ * \param font the font to specify a direction for.
+ * \param char_spacing the new spacing value.
+ * \returns true on success or false on failure; call SDL_GetError() for more
+ *          information.
+ *
+ * \threadsafety This function should be called on the thread that created the
+ *               font.
+ *
+ * \since This function is available since SDL_ttf 3.0.0.
```
slouken, line 990:
```suggestion
 * \since This function is available since SDL_ttf 3.4.0.
```

include/SDL3_ttf/SDL_ttf.h
```diff
@@ -969,6 +969,44 @@ extern SDL_DECLSPEC bool SDLCALL TTF_SetFontDirection(TTF_Font *font, TTF_Direct
...
+ *
+ * This defaults to 0 if it hasn't been set.
+ *
+ * \param font the font to query.
+ * \returns the character spacing in pixels.
+ *
+ * \threadsafety This function should be called on the thread that created the
+ *               font.
+ *
+ * \since This function is available since SDL_ttf 3.0.0.
```
slouken, line 1006:
```suggestion
 * \since This function is available since SDL_ttf 3.4.0.
```

slouken:
> I think 1 glyph isn't one character in all languages. This works for latin, but not for more complicated language arabic, etc. also ligature chars ?
> 
> HarfBuzz should do the letter-spacing .. maybe that's the current state: [typst/typst#6161](https://github.com/typst/typst/pull/6161)

Oh, that's a really good point. @ivan-mogilko, can you take a look at this?

slouken:
> @slouken , on a separate note, there was another mine PR (#538) opened few months ago made for SDL2 branch, but there was no update to it for a while. Should I also remake this on the main branch instead?

We're not planning to add new functions to the SDL2 API at this point.

Separately from that, I was concerned that people might be confused about which function they should call to get measurement information, and I haven't sat down and thoroughly thought it through. Could you provide a use case for this?

PR #579, opened 2025-08-13
require freetype version at least 2.8.0

sezero:
Current SDL3_ttf no longer builds against old freetype versions due
to uncontitional uses of FT_COLOR features available only in v2.5.1
and newer.

RedHat Enterprise Linux 8 (RHEL8) ships with version 2.9.1, whereas
Debian oldoldstable has 2.10.4. Therefore, setting our minimum vers
as 2.8 should satisfy everyone.

On the other hand, SDF still requires freetype >= 2.11.0 and, as it
seems, it is conditionalized in SDL3_ttf not resulting in any build
failures.

If accepted, I suggest cherry-picking it into release-3.2.x as well


madebr (approved):
If it was already broken, I'm fine with enforcing a minimum compatibility version.

slouken (approved):

PR #580, opened 2025-08-16
Made the installation instructions clearer

KristianNikolov07:


slouken:
I don't know that this is clearer, and the instructions are intended to be used by people who've downloaded the release archive instead of cloning via git.

PR #592, opened 2025-10-30
Correct migration guide for TTF_SizeText/TTF_SizeUTF8

gbgames:
According to the .cocci file, they both get replaced by TTF_GetStringSize(), which matches their signatures (aside from the return type) and makes it a simple rename.


slouken:
Merged, thanks!

PR #600, opened 2025-11-26
Fix memcpy misuse in TTF_DeleteTextString

mcendu:
AddressSanitizer aborts the program when calling `TTF_DeleteTextString`, which in turn calls `SDL_memcpy` instead of the correct `SDL_memmove`.

This commit fixes the abort under ASan.


slouken:
Merged, thanks!

PR #611, opened 2026-02-13
fixed: sdf character layout was misaligned

ProPuke:
The glyphs were misaligned when rendering sdf text (at least with `TTF_GetGPUTextDrawData()` data).
The bottom right of each glyph was seemingly not expanded to include the spread, shrinking each visible character.

Without SDF:
image

With SDF:
image

This change corrects that, bringing SDF back in line:
image

I haven't tested this with other rendering modes and setups (so it will need a sanity check)
But at a glance it looked like an incorrect typo (why would we be shrinking the dest rects?), and sure enough removing those 2 lines fixed alignment for me. So.. probably all good :P

### DIFF
--- a/src/SDL_ttf.c
+++ b/src/SDL_ttf.c
@@ -1504,8 +1504,6 @@ static bool Render_Line_TextEngine(TTF_Font *font, TTF_Direction direction, int 
             if (glyph_font->render_sdf) {
                 op->copy.dst.x -= DEFAULT_SDF_SPREAD;
                 op->copy.dst.y -= DEFAULT_SDF_SPREAD;
-                op->copy.dst.w -= DEFAULT_SDF_SPREAD;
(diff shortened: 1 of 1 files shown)


slouken:
I think that was actually supposed to be `+= 2*`, but maybe I'm wrong?

ProPuke:
> I think that was actually supposed to be `+= 2*`, but maybe I'm wrong?

I don't believe so. Leaving the dest rect the same size as the source is the only case that sized it correctly without shrink or stretch of the glyph.
I did test with just `+= DEFAULT_SDF_SPREAD` first, but that expanded the rects too much - 0 change kept things aligned.

But I have only tested with gpu rendering. It could be that things behave differently in other situations or through other render functions (although, if so, then we have other problems :D)

slouken:
I think you’re right, yours is the correct change, thanks!

slouken:
Merged, thanks!

PR #614, opened 2026-02-25
PSP: Fix crash when rendering solid, shaded or lcd text

sharkwouter:
On the Playstation Portable rendering text in any mode other than blended crashes the application. This only affects SDL2_ttf. It seems to be fixed in the SDL3 version. This bug has been around for a while [here](https://github.com/pspdev/psp-packages/issues/45), but it wasn't until recently that @isage found the fix that is found within this PR.


slouken:
Merged, thanks!

PR #618, opened 2026-03-19
include module support for either SDF or SVG

pmp-p:
because both use FT_Property_Set around https://github.com/libsdl-org/SDL_ttf/blob/a49130e3ecc0ad5a4aad12083e9aeab57ac34dbd/src/SDL_ttf.c#L1872


slouken:
Merged, thanks!

